PR #18, opened 2019-09-18
Fixed no-op case in common-ngx-settings-service

Proantagonist:
Removed some lines that lead to a no-op case when updating setting values in client local storage.

The operation was already being performed on setttings service state change.


Proantagonist:
Removed `updateStorage`. Local store service will require a more generalized implementation for setting local storage values.

PR #24, opened 2019-09-30
Eliminate a bunch of `any` types

bigfootjon:


apps/aggiemap/src/app/modules/components/accessibility/map-viewfinder/map-viewfinder.component.ts
```diff
@@ -115,7 +115,7 @@ export class MapViewfinderComponent implements OnInit, OnDestroy {
             this.moduleProvider.require(['Polygon']).then(([Polygon]: [esri.PolygonConstructor]) => {
               // Create a polygon that will be used to query the feature layer to get intersecting features
               const polygon = new Polygon({
-                rings: mapCoords
+                rings: [mapCoords]
```
Bigfootjon: Functional change

apps/aggiemap/src/app/modules/directives/accordion.directive.ts
```diff
@@ -67,7 +68,8 @@ export class AccordionDirective implements OnChanges, OnInit, OnDestroy {
     if (this.resize) {
       this.mutationObserver = new MutationObserver((mutations: MutationRecord[]) => {
         this.setAccordionHeight();
-      }).observe(this.el.nativeElement, {
+      });
+      this.mutationObserver.observe(this.el.nativeElement, {
```
Bigfootjon: Functional change,

`.observe(...)` returns void. So I think this is what this block of code was supposed to look like

libs/common/utils/geometry/esri/src/lib/common-utils-geometry-esri.ts
```diff
@@ -12,18 +12,17 @@ import { polygon as tPolygon } from '@turf/helpers';
  * @param feature Esri geometry
  */
 export function centroidFromGeometry(
-  feature: esri.Geometry | esri.Polygon | esri.Multipoint | esri.Point | esri.Polyline
-): Point;
-export function centroidFromGeometry(feature: any): Point {
-  if (feature.rings) {
+  feature: esri.Geometry | esri.Polygon | esri.Multipoint | esri.Point | esri.Polyline | Point
+): Point {
+  if ('rings' in feature) {
```
Bigfootjon: Not exactly a functional change, but this can be used in TypeScript to scope the properties of a type

libs/common/utils/geometry/generic/src/lib/common-utils-geometry-generic.ts
```diff
@@ -25,7 +26,7 @@ export function getGeolocation(asObservable?: any): Promise<any> | Observable<an
 
   if (asObservable) {
     return from(promise);
-  } else {
+  } else if (asObservable as false) {
```
Bigfootjon: TODO: remove this

tslint.json
```diff
@@ -37,6 +37,7 @@
     "no-unused-expression": true,
     "no-var-keyword": true,
     "object-literal-sort-keys": false,
+//    "no-any": true,
```
Bigfootjon: TODO Remove this

libs/common/utils/object/src/lib/common-utils-object.ts
```diff
@@ -9,18 +9,14 @@
...
   if (!lookup || !property) {
     return undefined;
   }
 
   const path = property.split('.');
   let value = JSON.parse(JSON.stringify(lookup));
 
-  if (path.length === 0) {
-    return undefined;
-  }
```
bigfootjon: According to [the spec](http://www.ecma-international.org/ecma-262/5.1/#sec-15.5.4.14) it is impossible for `String.split` to return an array with less than 1 element. In the case that the pattern (in this case: `'.'`) is not found then the entire string is returned.

Proantagonist: Good catch.

PR #29, opened 2019-10-28
Fix test cases

bigfootjon:
I was confused about how to invoke `inject` in Jest test cases. Below is the faulty method I was using before followed by the corrected method:

````typescript
// OLD and BAD:
it('should be created', () => {
  inject([EXAMPLEService], (service: EXAMPLEService) => {
    expect(service).toBeTruthy();
  });
});

// NEW and GOOD:
it('should be created', inject([EXAMPLEService], (service: EXAMPLEService) => {
  expect(service).toBeTruthy();
}));
````
However, it is not sufficient to just include the `inject` call from the testing module, the `TestBed` must be configured for each test:
````typescript
beforeEach(async(() => {
  TestBed.configureTestingModule({
    providers: [EXAMPLEService],
  }).compileComponents();
}));
````
This testing module must import all required dependencies for the module. The correct way to invoke `inject` actually runs the code (what a concept) and the executed lines get counted towards coverage


bigfootjon:
Increased test coverage 9%!

unlimited power

PR #33, opened 2019-11-07
Fix duplicate URL params causing visual glitch

bigfootjon:
When a building is specified multiple times in a URL param, the highlight "stacks" so that building is over-highlighted compared to other buildings. E.g. `bldg=ZACH,ZACH,ZACH,CHEN` CHEN has correct highlight, but ZACH has too much:

Screenshot from 2019-11-07 12-26-18

While I was fixing this, I noticed that `strip` isn't a function, so I changed that to `trim`, fixed a typo in a comment, and removed the "reveille-console-log works!" default comment


bigfootjon:
@nicholasjanzen thanks for reporting

PR #42, opened 2019-12-13
BikeShare Visualization Application 

ryanbilek:
Using AggieMap Trees as a base, the Bikeshare application is able to visualize the current locations of VeoRide bikes, and view concentrations and idle areas through heatmaps. Origin and destination heat maps allow for a user to view concentrations of where bikes are leaving and arriving. This can be changed with the date picker.

The layers are pulling data from urls defined by _apps/bikeshare/src/environments/definitions.ts_.

The origin and destination layers currently are unable to be reloaded, which translates to only data for the first date chosen is loaded when the origin/destination is selected. To get a different date, the application must be reloaded and a new date picked. 

The dates of the origin/destination layers are currently changed in _libs/maps/feature/layer-list/src/lib/services/layer-list.service.ts_ by appending a date argument to the url that the application is pulling data from. Using _layer-list.service_ to do this is bad practice. The layers should be changed in separate component within the application itself, which should currently interface with the map service to kill and remake the layers and leave the library alone.


libs/search/src/lib/services/search.service.spec.ts
```diff
@@ -69,3 +69,27 @@ describe('SearchResult', () => {
...
+          breadcrumbs: {
+            source: 'tests',
+            value: 'probably nothings'
+          }
+        }
+      ]
+    });
+    expect(search_result.features()).toEqual(['test', 'test2']);
+  });
+});
```
Proantagonist: Is this an accidental duplicate?

libs/ui-kits/ngx/branding/src/lib/reveille-console-log/reveille-console-log.component.ts
```diff
@@ -1,54 +1,54 @@
...
+     .oddy+-\`   \`/yddddh+.\`  \`--::/::-.\`  \`-+ydddhddddddy+-\`      -yddo\`   \`yhdhd-
+    :hdh+\`     \`ydddddddddyo/:..\`\`\`\`\`..:+oyhddhy/+dddho-\`          \`+ddo\`   :ddhds\`
+   /hddy:\`.+-   .:+ooooooyhdddddhhhhhdddddhhs+-\`/ddh+.               +ddo\`   oddddys+:.\`
+   -yds\`\`+hhyo-\`          ./shddddddddho+:-\` \`.odh+\`   \`\`\`\`\`\`         +dd+   \`hddddddhhyo\`
+   +dh\`\`syhhhhhs+-.-::.\`    \`-sddhshdh.   \`./shhdyooooosyyyyys:\`       /hd+   .+shhddddddy\`
+  \`hd+ -+ohhhhhhhhhhhhho-\`  \`/shds-.:/    :hhhhhhhhhysssyhdddddo        :hd/\`    \`-+hhhhy:
+  .dd+  :hhhhhhhhhhhhhhhhs/.\`  -oyhyo/.\`   \`..--:-.\`     .+ddy+\`         -yhy+-     :dh-
+  \`hhy\` -yyhs//oyyyhy+++o++/.    \`.:+ydys-\`\`\`\`\`\`\`\`\`\`\`\`\`\`\`.ohh-            \`:shhs+//+shh/
+   /hho.\`\`---.  \`.---\` .-\`\`\` \`/+oosyhhhdhyyyyyyyyyyyyyyyyhh+.                .:/+osso/.
+    :yhhso/::/++ossoo//+hhysssydddoo+/::/+++/:::----------\`
```
Proantagonist: Intentional?

.angulardoc.json
```diff
@@ -0,0 +1,4 @@
+{
+  "repoId": "eeef0326-152b-4dee-bcd0-776e21b7c84e",
+  "lastSync": 0
+}
```
Proantagonist: This appears to be from a VSCode extension. Should probably stay on the client-machine, since it's not a configuration for any dev dependency.

apps/bikeshare/src/app/components/date-picker/date.component.scss
```diff
@@ -0,0 +1 @@
+@import "~ng-pick-datetime/assets/style/picker.min.css";
```
Proantagonist: Remove this import declaration. @import basically pastes the entirely of the target CSS file in here which means many declarations will result in a bloated output. On top of that, due to Angular's view encapsulation these styles will not apply to the date picker component nested inside of this component.

Import these styles once in the global styles.scss file.

apps/bikeshare/src/app/components/map/map.component.scss
```diff
@@ -0,0 +1,6 @@
+@import 'libs/sass/mixins';
+@import "~ng-pick-datetime/assets/style/picker.min.css";
```
Proantagonist: Remove this import declaration. @import basically pastes the entirely of the target CSS file in here which means many declarations will result in a bloated output. On top of that, due to Angular's view encapsulation these styles will not apply to the date picker component nested inside of this component.

Import these styles once in the global styles.scss file.

ryanbilek: Resolved via c7c8092.

apps/bikeshare/src/app/components/date-picker/date.component.scss
ryanbilek: Resolved via c7c8092.

libs/ui-kits/ngx/branding/src/lib/reveille-console-log/reveille-console-log.component.ts
ryanbilek: Removed and resolved in 0c29bdc.

.angulardoc.json
ryanbilek: Removed and resolved in cf52a2b1f98c4507f33dd1b7171104636094b829.

libs/search/src/lib/services/search.service.spec.ts
ryanbilek: Removed and resolved in a0432d5.

PR #49, opened 2020-06-04
Jorge's Covid Time Map

jorge-sepulveda:
Created a Maps module and a timemap component. This is an attempt to bring the existing timemap from personal github to our monorepo.


libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.html
```diff
@@ -0,0 +1,50 @@
+<section>
```
Proantagonist, line 1: Get rid of this `section`. It adds a margin to the top and bottom of the map and it's not particularly meaningful in this page.

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.scss
```diff
@@ -0,0 +1,163 @@
...
+    @include align-items(center);
+    width: 37.5%;
+    margin: 3rem 2rem;
+
+    .product-tiles_tile__identity {
+      width: 4.5rem;
+      margin-bottom: 1.25rem;
+    }
+  }
+}
```
Proantagonist, line 21: Remove unused CSS

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.scss
```diff
@@ -0,0 +1,163 @@
...
+
+#geocoding-graphic {
+  height: 20rem;
+  object-fit: cover;
+}
```
Proantagonist, line 26: Remove unused CSS

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.scss
```diff
@@ -0,0 +1,163 @@
...
+
+/* Rounded sliders */
+.slider.round {
+  border-radius: 34px;
+}
+
+.slider.round:before {
+  border-radius: 50%;
+}
```
Proantagonist, line 123: This one is a matter of consistency. My intent here is not to prevent you from having any creative freedom, however it's worth considering that these styles are not usable anywhere else in the application/monorepo in their current state. Either pull them out into the main application styles so they can be re-used on the other remaining maps or consider adopting the pre-styled form input components that are used in other parts of this application. I'm also perfectly OK with having a separate "skin" for input by creating a new "view" that extends the current checkbox input with a different set of styles. In this way, the styles would be bound to a checkbox component, and not a consuming component. 

Whatever you want to do here is up to you, but the overall goal should be consistency and as much DRY-ness as possible.

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.scss
```diff
@@ -0,0 +1,163 @@
...
+.legend {
+	background-color: #fff;
+	border-radius: 3px;
+	bottom: 30px;
+	box-shadow: 0 1px 2px rgba(0, 0, 0, 0.1);
+	font: 12px/20px 'Helvetica Neue', Arial, Helvetica, sans-serif;
+	padding: 10px;
+	position: absolute;
+	z-index: 1;
+}
```
Proantagonist, line 137: Is the goal here to have the legend at the bottom of the sidebar, or at the bottom right of the map?

It's currently absolutely positioned relative to the sidebar, which has no inherited "relative" position, so it's effectively absolutely positioned relative to the viewport.

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.ts
```diff
@@ -0,0 +1,293 @@
+import { Component, OnInit } from '@angular/core';
+import { HttpClient } from '@angular/common/http';
+import { MapboxMapService } from '@tamu-gisc/maps/mapbox';
+import { Moment } from 'moment';
+import { forkJoin } from 'rxjs';
```
Proantagonist, line 5: Import ordering:

- First party libraries (angular, nestjs, etc)
- Third party NPM libraries (moment)
- Monorepo scoped imports (@tamu-gisc)
- Relative file imports
- Type imports

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.ts
```diff
@@ -0,0 +1,293 @@
...
+  selector: 'tamu-gisc-time-map',
+  templateUrl: './time-map.component.html',
+  styleUrls: ['./time-map.component.scss']
+})
+export class TimeMapComponent implements OnInit {
+  constructor(private mapService: MapboxMapService, private http: HttpClient) {}
+
+  private maxDate: string;
+  private mortalButtonToggled: boolean;
+  private stateButtonToggle: boolean;
```
Proantagonist, line 17: Default value can be assigned here (`false`)

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.ts
```diff
@@ -0,0 +1,293 @@
...
+        'state-label'
+      );
+
+      map.on('mousemove', 'covid-county', function(e) {
+        const feature = e.features[0];
+        const selectedCounty = initObject.countyData[initObject.selectedDate].filter(
+          (county) => county.fips === feature.properties.fips
+        );
+
+        document.getElementById('info-box').innerHTML =
```
Proantagonist, line 89: Do not use direct `document` methods to access DOM elements.

Prefer the use of the `@ViewChild` decorator using template references to query the DOM.

```html
<div #reference>
    <!-- Stuff goes here -->
</div>
```

```js
@ViewChild('reference') public container: ElementRef;

this.container.nativeElement // <-- This is your element.
```

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.ts
```diff
@@ -0,0 +1,293 @@
...
+    this.stateButtonToggle = !this.stateButtonToggle;
+    this.switchLayers();
+  }
+
+  public mortalityToggle(toggleButton) {
+    this.mortalButtonToggled = !this.mortalButtonToggled;
+    this.mortalButtonToggled ? this.drawDeathMap() : this.drawCasesMap();
+    if(this.mortalButtonToggled){
+      this.drawDeathMap()
+      document.getElementById('cases-legend').style.display = 'none';
```
Proantagonist, line 276: Same thing here with using `document`.

Additionally, avoid setting styles in code. Use CSS classes or template directives like `NgClass`, `NgStyle`, or other structural directives such as `NgIf`.

Since it seems all you want to do is hide/show something, use `NgIf`

```html
    <div *ngIf="mortalButtonToggled ">
         <!-- The ngIf directive will evaluate the mortalButtonToggled value and hide/show based on its coerced boolean value. This has the benefit of completely removing the element from the DOM when the expression evaluates to false. -->
    </div>
```

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.ts
```diff
@@ -0,0 +1,293 @@
...
+    }
+    else{
+      this.drawCasesMap();
+      document.getElementById('cases-legend').style.display = 'block';
+			document.getElementById('mortality-legend').style.display = 'none';
+    }
+  }
+
+  public switchLayers() {
+    this.mapService.loaded.subscribe((map) => {
```
Proantagonist, line 287: Do not create new subscriptions for every layer switch. `loaded` is a long-lived observable and the others subscriptions you create will not close automatically.

The `mapService` exposes the `map` instance, so use that.

```js
this.mapService.map.setLayoutProperty(...)
```

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.ts
```diff
@@ -0,0 +1,293 @@
...
+          : number > 1
+          ? '#dadaeb'
+          : '#f2f0f7';
+      countyExpression.push(row['fips'], color);
+    });
+
+    stateExpression.push('rgba(255,255,255,1)');
+    countyExpression.push('rgba(255,255,255,1)');
+
+    this.mapService.loaded.subscribe((map) => {
```
Proantagonist, line 260: `loaded` is a long-lived observable. Schedule the subscription close on component destroy using the `takeUntil` operator

https://medium.com/angular-in-depth/the-best-way-to-unsubscribe-rxjs-observable-in-the-angular-applications-d8f9aa42f6a0

Or limit the number of emissions with the `take` rxjs operator.

https://www.learnrxjs.io/learn-rxjs/operators/filtering/take

libs/geoservices/covid/src/lib/pages/maps/maps.module.ts
```diff
@@ -0,0 +1,35 @@
+import { NgModule } from '@angular/core';
+import { CommonModule } from '@angular/common';
+import { RouterModule, Routes } from '@angular/router';
+
+import { MapsComponent } from './maps.component';
+import { TimeMapComponent } from './components/time-map/time-map.component';
+import { MapsMapboxModule } from '@tamu-gisc/maps/mapbox';
+import { HttpClientModule } from '@angular/common/http';
+import { UILayoutModule } from '@tamu-gisc/ui-kits/ngx/layout';
```
Proantagonist, line 9: Import ordering

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.html
```diff
@@ -0,0 +1,50 @@
...
+  <div id="map-section">
+    <tamu-gisc-drawer [width]="'25rem'" [position]="'left'">
+      <div class="content">
+        <h3>Map Controls</h3>
+        <label class="switch">
+          <input type="checkbox" id="csToggle" (click)="stateToggle(this)">
+          <span class="slider round"></span>
+        </label>
```
Proantagonist, line 9: Label should not wrap the input element.
```html
<input type="checkbox" name="county-state-toggle">
<label for="county-state-toggle>Counties/State</label>
```

libs/geoservices/covid/src/lib/geoservices-covid.module.ts
```diff
@@ -22,6 +22,10 @@ const routes: Routes = [
         path: 'services',
```
Proantagonist, line 22: Use formatter on all files.

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.html
```diff
@@ -0,0 +1,50 @@
...
+    <tamu-gisc-mapbox-map
+      token="pk.eyJ1Ijoid2hhdHppdG9veWEiLCJhIjoiY2poNzI3M2M0MDNnbjMzcWVtYmZnNXExNiJ9.K-nM-JaZnZvAfutkLQitbg"
+      [center]="['-98', '38.88']">
+    </tamu-gisc-mapbox-map>
+    <div class = 'right-sidebar'>
+      <h3>COVID-19 Data</h3>
+      <div id= "info-box">
+        <p>Hover over the map to see info</p>
+      </div>
+      <div id="cases-legend" class="legend" style="display: block;">
```
Proantagonist, line 28: No inline styles.

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.ts
```diff
@@ -0,0 +1,293 @@
...
+          'Infection Rate: ' +
+          selectedCounty[0]['infection_rate'].toFixed(2) +
+          '/100,000 People</br>' +
+          'Deaths: ' +
+          selectedCounty[0]['deaths'] +
+          '</br>' +
+          'Mortality Rate: ' +
+          selectedCounty[0]['death_rate'].toFixed(2) +
+          '/100,000 People';
+      });
```
Proantagonist, line 108: Separation of concerns. Do not compose templates in code, use actual templates.

Create a public member, update that, and use it in the template:


```js

public infoBoxModel;

this.infoBoxModel =  {
    county: feature.properties.Name,
    pop: feature.properties.POPESTIMATE2019,
    ...,
    ....
}
```

In template:

```html
    <p>County: {{infoBoxModel.county}}</p>
    <p>Population: {{infoBoxModel.pop}}</p>
```

Avoid the use of `<br>` tags for spacing.

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.ts
```diff
@@ -0,0 +1,293 @@
...
+  styleUrls: ['./time-map.component.scss']
+})
+export class TimeMapComponent implements OnInit {
+  constructor(private mapService: MapboxMapService, private http: HttpClient) {}
+
+  private maxDate: string;
+  private mortalButtonToggled: boolean;
+  private stateButtonToggle: boolean;
+  private selectedDate: string;
+  private stateData: object;
```
Proantagonist, line 19: Avoid the use of the basic `object` type.

The `object` type will not allow you to access members that are not part of the base object prototype because it assumes it's just that. It has no idea if the object is an array, and you will not receive any help from the editor with method suggestions.

```js
private stateData: object;

this.stateData.someProperty = "some value"; // Property 'someProperty' does not exist on type 'object'

this.stateData.forEach((s, i, a) => {
  // Logic
}) //Property 'forEach' does not exist on type 'object'
```

If you know the data structure, create interfaces. This will provide type-checking through your application and provide suggestions in template.

```js
interface IStateRecord {
    name: string;
    population: number;
}
```

```js
private stateData: Array<IStateRecord>;
```

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.html
```diff
@@ -0,0 +1,50 @@
...
+      <div class="content">
+        <h3>Map Controls</h3>
+        <label class="switch">
+          <input type="checkbox" id="csToggle" (click)="stateToggle(this)">
+          <span class="slider round"></span>
+        </label>
+        Counties/State
+        <br>
+        <label class="switch">
+          <input type="checkbox" id="mortalToggle" (click)="mortalityToggle(this)">
```
Proantagonist, line 13: Use `$event` instead of `this`. 

`this` refers to the component class instance.

`$event` refers to the event object.

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.ts
```diff
@@ -0,0 +1,293 @@
...
+  private countyData: object;
+
+  public ngOnInit() {
+    const initObject = this;
```
Proantagonist, line 23: Not necessary. `this` is available within the component class at any point during it's life cycle.

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.ts
```diff
@@ -0,0 +1,293 @@
...
+            'line-color': '#000000',
+            'line-width': 1
+          }
+        },
+        'state-label'
+      );
+
+      map.on('mousemove', 'covid-county', function(e) {
+        const feature = e.features[0];
+        const selectedCounty = initObject.countyData[initObject.selectedDate].filter(
```
Proantagonist, line 85: Use arrow functions unless you require a new scope inside the function.

This will also remove the need to bind `initObject` because arrow functions are lexically bound, meaning `this` will refer to the parents scope (component class).

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.scss
jorge-sepulveda: It can be either or, the legend fits nicely in the sidebar so I would choose to keep it there unless you prefer it in the map

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.html
jorge-sepulveda: removing "this" and the parameter in the function declaration since it's not needed to perform the layer switch

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.html
jorge-sepulveda: removed it

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.scss
jorge-sepulveda: removed

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.scss
jorge-sepulveda: I'll remove this and use the checkbox mentioned later here

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.ts
jorge-sepulveda: ordered

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.ts
jorge-sepulveda: resolved

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.ts
jorge-sepulveda: removed usage of direct document api access

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.ts
jorge-sepulveda: used ngif on the legend element

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.ts
jorge-sepulveda: removed the subscription

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.scss
Proantagonist: I'm OK with either. These styles currently dock the legend at the bottom right of the screen, not necessarily inside the sidebar. At higher resolutions, it shows on top of the footer.

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.ts
jorge-sepulveda: used `take(1)` wherever necessary throughout the code.

libs/geoservices/covid/src/lib/pages/maps/maps.module.ts
jorge-sepulveda: ordered

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.html
jorge-sepulveda: used this checkbox setup and removed the old switch and it's css

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.html
jorge-sepulveda: gave these id and assigned them a color in the css file

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.ts
jorge-sepulveda: removed initObject and made the mouse moves into arrow functions

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.ts
jorge-sepulveda: removed it

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.ts
jorge-sepulveda: created interfaces for the county and state data.

jorge-sepulveda: had some trouble with the array implementation so I'm just removing the 'object' type from the declaration

jorge-sepulveda:
Got all of these resolved except for the interface implementation of the state and county record. I'm guessing there's a small thing I'm overlooking. I will push my changes so you can review the other fixes in the meantime.

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.html
```diff
@@ -0,0 +1,50 @@
...
+    <div #infobox *ngIf="(infoBoxModel | async) as info">
+      <p *ngIf="!stateButtonToggle">County: {{info.name}}</p>
+      <p *ngIf="stateButtonToggle">State: {{info.name}}</p>
+      <p>Population: {{info.pop}}</p>
+      <p>Cases: {{info.cases}}</p>
+      <p>Infection Rate: {{info.infectionRate.toFixed(2)}}</p>
+      <p>Deaths: {{info.deaths}}</p>
+      <p>Mortality Rate: {{info.deathRate.toFixed(2)}}</p>
+    </div>
+    <div class="legend">
```
Proantagonist, line 29: Should this have an `NgIf` or does it show all the time, even when the mortality rate layer is enabled?

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.ts
```diff
@@ -0,0 +1,276 @@
...
+})
+export class TimeMapComponent implements OnInit {
+  constructor(private mapService: MapboxMapService, private http: HttpClient) {}
+
+  private maxDate: string;
+  private selectedDate: string;
+  private stateData: Array<StateRecord>;
+  private countyData: Array<CountyRecord>;
+
+  @ViewChild('infobox', { static: false }) public infobox: ElementRef;
```
Proantagonist, line 51: Since you're using `infoBoxModel` to render values in the template, this element reference might not be necessary.

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.html
jorge-sepulveda: When the mortality rate is enabled, this dissapears. I can add ngIf for consistency with the other legend.

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.ts
jorge-sepulveda: noted. removed the element reference

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.html
```diff
@@ -0,0 +1,50 @@
+<div id="map-section">
+  <tamu-gisc-drawer [width]="'20rem'" [position]="'left'">
+    <div class="content">
+      <h3>Map Controls</h3>
+      <input type="checkbox" name="county-state-toggle" (click)="stateToggle()">
+      <label for="county-state-toggle">Counties/State</label>
```
Proantagonist, line 6: Feel free to use `<tamu-gisc-checkbox>` component from `UIFormsModule`.

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.html
```diff
@@ -0,0 +1,50 @@
...
+      <br>
+      <input type="checkbox" name="infection-mortality-toggle" (click)="mortalityToggle()">
+      <label for="county-state-toggle">Infection Rate/Mortality Rate</label>
+      <input type="date" id="datepicker" value="2020-06-06" max="2020-06-06" min="2020-01-21">
```
Proantagonist, line 10: We also have `tamu-gisc-date-time-picker` but it hasn't been fully tested. So your call if you want to spend time trying that one out, definitely not necessary.

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.html
jorge-sepulveda: done

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.scss
```diff
@@ -0,0 +1,101 @@
...
+//legend start
+.legend {
+  background-color: #fff;
+  border-radius: 3px;
+  bottom: 7rem;
+  box-shadow: 0 1px 2px rgba(0, 0, 0, 0.1);
+  padding: 10px;
+  position: absolute;
+  z-index: 1;
+}
```
Proantagonist, line 37: Legend is still positioned absolutely relative to the viewport (overlayed on top of the footer).

Add `position: relative;` to the right sidebar place it relative to the sidebar.

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.html
jorge-sepulveda: I'm going to remove the button and the datepicker for now because there is no logic attached to them at the moment. when I add the date feature, I can look into the date-time-picker component.

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.ts
```diff
@@ -0,0 +1,276 @@
...
+      countyExpression.push(row['fips'], color);
+    });
+
+    stateExpression.push('rgba(255,255,255,1)');
+    countyExpression.push('rgba(255,255,255,1)');
+
+    this.mapService.loaded.pipe(take(1)).subscribe((map) => {
+      map.setPaintProperty('covid-state', 'fill-color', stateExpression);
+      map.setPaintProperty('covid-county', 'fill-color', countyExpression);
+    });
```
Proantagonist, line 211: You can also use the `map` instance directly from the `mapService` much like you did in the `switchLayers` toggle.

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.ts
```diff
@@ -0,0 +1,276 @@
...
```
Proantagonist, line 258: Same here, consider eliminating the subscription altogether and do like you did in `switchLayers` method.

libs/geoservices/covid/src/lib/pages/maps/maps.component.html
```diff
@@ -0,0 +1 @@
+<router-outlet></router-outlet>
```
Proantagonist, line 1: Prettier should be adding newlines at the end of a file?

libs/geoservices/covid/src/lib/pages/maps/maps.component.ts
```diff
@@ -0,0 +1,15 @@
...
+@Component({
+  selector: 'tamu-gisc-maps',
+  templateUrl: './maps.component.html',
+  styleUrls: ['./maps.component.scss']
+})
+export class MapsComponent implements OnInit {
+
+  constructor() { }
+
+  ngOnInit() {
```
Proantagonist, line 12: All class members should have member access defined `public`, `private`, `protected`.

If you hit F1 in VSCode, and type out `Fix All`, VSCode will automatically append the most appropriate access type based on class usage. Defining these will set constraints when extending classes or when certain properties should not be used in templates.

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.ts
jorge-sepulveda: resolved

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.ts
jorge-sepulveda: resolved

libs/geoservices/covid/src/lib/pages/maps/maps.module.ts
```diff
@@ -0,0 +1,37 @@
...
+        component: TimeMapComponent
+      }
+    ]
+  }
+]
+
+@NgModule({
+  declarations: [MapsComponent, TimeMapComponent],
+  imports: [
+    CommonModule, RouterModule.forChild(routes), MapsMapboxModule, HttpClientModule,UILayoutModule, UIFormsModule
```
Proantagonist, line 34: Please remember to trigger your formatter on every file changed. Space should be found between `HttpClientModule` and `UILayoutModule`. Committing and pushing these files consistently formatted will prevent white space changes on files across contributors.

libs/geoservices/covid/src/lib/pages/maps/maps.component.html
jorge-sepulveda: forgot to run format document on code. I'll set it to run on save.

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.ts
```diff
@@ -0,0 +1,276 @@
...
+}
+
+interface IInfoBox {
+  name: string;
+  pop: number;
+  cases: number;
+  deaths: number;
+  infectionRate: number;
+  deathRate: number;
+}
```
Proantagonist, line 36: You can take this a step further and make a common interface which these extend.

`pop`, `cases`, `deaths`, `infectionRate`, and `deathRate` appear to be shared properties with the same types.

You can do this, for example:

```js
interface BaseRecord {
    pop: number;
    cases: number;
    deaths:  number;
    infectionRate: number;
    deathRate: number;
}
```

And then:

```js
interface StateRecord extends BaseRecord {
    STATE: string;
}
```
Lastly, in places where you have interfaces inside a decorated class file, please include the types/interfaces at the end. The decorator should be displayed immediately after imports.

Typescript doesn't care that interfaces or types are defined before using them.

libs/geoservices/covid/src/lib/pages/maps/maps.module.ts
jorge-sepulveda: enabled the formatter, the next push will reflect Prettier formatting.

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.ts
jorge-sepulveda: I'll make this fix. Also, does it matter if we declare the interfaces in a seperate ts file and import it to the component implementation or should we keep those at the bottom?

Proantagonist: You can definitely split it up into a different file. Normally, the data interfaces are stored alongside the data access API services/controllers, however the data for this map isn't generated by anything in the mono-repo so I'm not too sure where it would be saved. If the same interfaces will be used by the other maps you plan to add, then at that point it would make a whole lot more sense to decouple them from the component altogether.

Perhaps in:

`libs/geoservices/covid/src/lib/pages/maps/types`?

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.scss
jorge-sepulveda: added

libs/geoservices/covid/src/lib/pages/maps/components/time-map/time-map.component.scss
```diff
@@ -40,118 +19,39 @@
...
-	padding: 10px;
-	position: absolute;
-	z-index: 1;
+  background-color: #fff;
+  border-radius: 3px;
+  bottom: 1rem;
+  box-shadow: 0 1px 2px rgba(0, 0, 0, 0.1);
+  padding: 10px;
+  position: absolute;
+  z-index: 1;
```
Proantagonist: Since the legend is part of the sidebar, does it make sense for it to have a drop shadow or even a background color?

If you're feeling crazy it could also be done with a flex layout. 

Sidebar: flex, column, justify space between. This would force the info and the legend on the opposite ends of the sidebar.

PR #50, opened 2020-06-12
Sites api addition

jorge-sepulveda:
added functionality to the stats endpoint in order to use the endpoint to create a more effective map for the covid data collection team.

### DIFF
--- a/libs/covid/data-api/src/lib/modules/counties/counties.service.ts
+++ b/libs/covid/data-api/src/lib/modules/counties/counties.service.ts
@@ -44,7 +44,7 @@ export class CountiesService extends BaseService<County> {
     });
   }
 
-  public async getCountyStats() {
+  public async getCountyStats(): Promise<CountyStats> {
     const counties = (await this.repo
       .createQueryBuilder('county')
       .leftJoinAndMapMany('county.claims', CountyClaim, 'claim', 'claim.countyFips = county.countyFips')
--- a/libs/covid/data-api/src/lib/modules/counties/counties.service.ts
+++ b/libs/covid/data-api/src/lib/modules/counties/counties.service.ts
@@ -2,7 +2,7 @@ import { Injectable } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
 import { Repository, Like } from 'typeorm';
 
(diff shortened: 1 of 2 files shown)


libs/covid/data-api/src/lib/modules/lockdowns/lockdowns.service.ts
```diff
@@ -343,3 +351,21 @@ export class LockdownsService extends BaseService<Lockdown> {
...
+}
+
+interface LockdownStat {
+  claims?: number;
+  lockdowns?: number;
+}
+
+export interface LockdownStats {
+  [key: string]: LockdownStats;
+}
```
Proantagonist, line 371: These don't appear to be used anywhere. They have structure to the interfaces in the counties service.

Clean up any unused import statements.

libs/covid/data-api/src/lib/modules/counties/counties.service.ts
```diff
@@ -2,7 +2,7 @@ import { Injectable } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
 import { Repository, Like } from 'typeorm';
 
-import { County, CountyClaim, StatusType, EntityStatus, TestingSite } from '@tamu-gisc/covid/common/entities';
+import { County, CountyClaim, StatusType, EntityStatus, TestingSite, Lockdown } from '@tamu-gisc/covid/common/entities';
```
Proantagonist, line 5: Cleanup unused imports

libs/covid/data-api/src/lib/modules/lockdowns/lockdowns.service.ts
jorge-sepulveda: deleted it

libs/covid/data-api/src/lib/modules/counties/counties.service.ts
jorge-sepulveda: Lockdown  is used in the ClaimWithData interface

Proantagonist: `StatusType` is the odd one left out. That was my doing I think.

PR #51, opened 2020-06-15
Counties api mod

jorge-sepulveda:
Modifying the "counties/stats/" endpoint to return lockdown information. This is so the map that track lockdowns can easily color code the counties based on lockdowns timestamps.

### DIFF
--- a/libs/covid/data-api/src/lib/modules/counties/counties.service.ts
+++ b/libs/covid/data-api/src/lib/modules/counties/counties.service.ts
@@ -58,12 +58,15 @@ export class CountiesService extends BaseService<County> {
       const t: CountyStat = (acc[countyString] = {});
 
       t.claims = curr.claims.length;
+
       t.sites = curr.claims.reduce((a, c) => {
         return a + c.sites.length;
       }, 0);
+
       t.lockdowns = curr.claims.reduce((a, c) => {
         return a + c.lockdowns.length;
       }, 0);
+
       t.lockdownInfo = curr.claims.map((c) => {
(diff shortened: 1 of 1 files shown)


libs/covid/data-api/src/lib/modules/counties/counties.service.ts
```diff
@@ -64,6 +64,20 @@ export class CountiesService extends BaseService<County> {
...
+      t.lockdownInfo = curr.claims.map((c) => {
+        if (c.lockdowns[0] === undefined) {
+          return null;
+        } else {
+          const lockdownStat: LockdownStat = {
+            updated: c.lockdowns[0].updated,
+
+            created: c.lockdowns[0].created,
+
+            guid: c.lockdowns[0].guid
```
Proantagonist, line 76: Remove spacing. 

Per our conversation, newlines should be placed between the `t` property assignments or before/after any thing beyond primitive value assignments (e.g. functions, objects, arrays), not between a list of any kind.

```js
      t.claims = { ... };

      t.sites = { ... };
      
      t.lockdowns = { ... };

      t.lockdownInfo = { ... };
```

Instead of:


```js
      t.claims = { ... };
      t.sites = { ... };   
      t.lockdowns = { ... };
      t.lockdownInfo = { ... };
```

jorge-sepulveda: forgot to add new lines on the county stat object assignments. only did that for the lockdown info assignments at first. fixed it now.

PR #52, opened 2020-06-19
Lockdown map component

jorge-sepulveda:
Addition of a lockdown map for covid lockdown data collectors.


libs/geoservices/covid/src/lib/pages/maps/components/lockdown-map/lockdown-map.component.ts
```diff
@@ -0,0 +1,258 @@
...
+          'source-layer': 'state-lines',
+          paint: {
+            'line-color': '#000000',
+            'line-width': 1
+          }
+        },
+        'state-label'
+      );
+
+      map.on('mousemove', 'counties-lockdowns', (e) => {
```
Proantagonist, line 79: Not a showstopper: If you can get the mapbox-gl types working on this library, if they aren't there yet already, try and type all of the things. Event objects can be particularly frustrating to work with them since there's so many of them and in this case should not be confused with native mouse events (?).

libs/geoservices/covid/src/lib/pages/maps/components/lockdown-map/lockdown-map.component.ts
```diff
@@ -0,0 +1,258 @@
...
+
+      map.on('click', 'counties-lockdowns', (e) => {
+        const feature = e.features[0];
+        const hoveredCounty = this.statData.filter((county) => county.fips === feature.properties.fips);
+        const updatedDate: Date = hoveredCounty[0]['lockdownInfo'][0]
+          ? hoveredCounty[0]['lockdownInfo'][0]['updated']
+          : null;
+        const createdDate: Date = hoveredCounty[0]['lockdownInfo'][0]
+          ? hoveredCounty[0]['lockdownInfo'][0]['created']
+          : null;
```
Proantagonist, line 94: This is one of the reasons why it event types should be used. The `click` event *should* be a generic, which would enable typescript to know what kind of features are coming back in the click event object which in turn allows you to use direct property accessors with dot notation (strongly typed) instead of instead of bracket notation (weakly typed, explicit `any`).

In the event that lockdown info interface changed, you'd see these as errors at build-time instead of run-time.

libs/geoservices/covid/src/lib/pages/maps/components/lockdown-map/lockdown-map.component.ts
```diff
@@ -0,0 +1,258 @@
...
+      });
+
+      map.dragRotate.disable();
+    });
+  }
+
+  public getFipsLockdownData(fipsCode: string) {
+    const lockdownUrl = 'https://nodes.geoservices.tamu.edu/api/covid/lockdowns/active/county/' + fipsCode;
+    const requests = forkJoin([this.http.get(lockdownUrl)]);
+    requests.pipe(take(1)).subscribe(([lockdownData]) => {
```
Proantagonist, line 116: `httpClient`'s http methods are self-completing, so the `take` operator is not necessary.

libs/geoservices/covid/src/lib/pages/maps/components/lockdown-map/lockdown-map.component.ts
```diff
@@ -0,0 +1,258 @@
...
+      console.log(lockdownData);
+      return lockdownData;
+    });
+  }
+
+  public getStats(): void {
+    const statsUrl = 'https://nodes.geoservices.tamu.edu/api/covid/counties/stats';
+    const requests = forkJoin([this.http.get(statsUrl)]);
+    requests.pipe(take(1)).subscribe(([recievedData]) => {
```
Proantagonist, line 125: Same here with the `take` operator.

libs/geoservices/covid/src/lib/pages/maps/components/lockdown-map/lockdown-map.component.ts
```diff
@@ -0,0 +1,258 @@
...
+      const statToAdd: StatRecord = {
+        fips: keyIndex,
+        state: keyIndex.substring(0, 2),
+        fipsNum: parseInt(keyIndex, 10),
+        claims: this.unformattedData[keyIndex]['claims'],
+        sites: this.unformattedData[keyIndex]['sites'],
+        lockdowns: this.unformattedData[keyIndex]['lockdowns'],
+        lockdownInfo: lockdownRecords
+      };
+      this.statData.push(statToAdd);
```
Proantagonist, line 158: Set `this.statData` to be this whole operation instead of pushing data to it, using a `map` method instead of a `forEach`.

`forEach` should only need to be used sparingly, when you don't actually want to mutate any data and instead want to call an action against some data. For example, triggering events.

`map` should be used when you have some data and want different data than the input, based on the input.

libs/geoservices/covid/src/lib/pages/maps/components/lockdown-map/lockdown-map.component.ts
```diff
@@ -0,0 +1,258 @@
...
+      this.unformattedData = recievedData;
+      this.formatData();
+      this.drawClaimsMap();
+      this.drawLockdownMap();
+    });
+  }
+
+  public formatData(): void {
+    this.statData = new Array();
+    let lockdownRecords = new Array();
```
Proantagonist, line 135: Be mindful of variable scoping. `lockdownRecords` is only used within the `forEach`, so it makes more sense to declare, initialize, and re-assign values inside that block. They will both be functionally the same, however. It's just one of those "best practices" things.

libs/geoservices/covid/src/lib/pages/maps/components/lockdown-map/lockdown-map.component.ts
```diff
@@ -0,0 +1,258 @@
...
+    Object.keys(this.unformattedData).forEach((keyIndex) => {
```
Proantagonist, line 136: Perhaps `Object.entries` is more appropriate? It seems like you're using indexes to access a parent object value and aren't accessing +1 or -1. 

`entries` provides both the key and value in the same scope, reducing the dependency on index values.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries

libs/geoservices/covid/src/lib/pages/maps/components/lockdown-map/lockdown-map.component.ts
```diff
@@ -0,0 +1,258 @@
...
+        return item !== null;
+      });
+      lockdownRecords = lockdownRecords.map((item) => {
+        const lockdownStat: LockdownRecord = {
+          updated: item.updated,
+          created: item.created,
+          guid: item.guid
+        };
+        return lockdownStat;
+      });
```
Proantagonist, line 147: You can chain these the `map` calls.

Example:

```js
const someArray = [0, 2, 3, undefined, 4, 5, 6];

console.log(someArray.filter(n => n !== undefined).map(n => n + 1));

// 1, 3, 4, 5, 6, 7
```

libs/geoservices/covid/src/lib/pages/maps/components/lockdown-map/lockdown-map.component.ts
Proantagonist: Marked resolved accidentally

libs/geoservices/covid/src/lib/pages/maps/components/lockdown-map/lockdown-map.component.ts
```diff
@@ -0,0 +1,258 @@
...
+        });
+      });
+
+      map.dragRotate.disable();
+    });
+  }
+
+  public getFipsLockdownData(fipsCode: string) {
+    const lockdownUrl = 'https://nodes.geoservices.tamu.edu/api/covid/lockdowns/active/county/' + fipsCode;
+    const requests = forkJoin([this.http.get(lockdownUrl)]);
```
Proantagonist, line 115: Separation of concerns.

As you make heavier user of the http library, you've probably noticed you need to prepare the URL before sending it out the request. Should the host or endpoint change, it would require going into these components and making several updates.

Separate the http access into a service which can be injected into this component and consumed here. In this way, you abstract out the URL, error handling, and any data transformation/cleaning up, before it makes it to the component, on top of having to make updates in a single place.

There is already a data-access library for this covid project and there are several exported services for various resources relative to this project. The library scope is:

'@tamu-gisc/geoservices/data-access'

This particular resource call is already abstracted as `LockdownsService.getLockdownForCounty`.

libs/geoservices/covid/src/lib/pages/maps/components/lockdown-map/lockdown-map.component.html
```diff
@@ -0,0 +1,45 @@
...
+    <div #infobox *ngIf="(infoBoxModel | async) as info">
+      <h4>Lockdown Info</h4>
+      <p>County: {{info.name}}</p>
+      <p>FIPS: {{info.fips}}</p>
+      <p>Claims: {{info.claims}}</p>
+      <p>Lockdowns: {{info.lockdowns}}</p>
+      <p>Updated: {{info.updated}}</p>
+      <a [routerLink]="['/admin/claims']" [queryParams]='{state: info.state, county: info.fipsNum}' target="_blank">
+        Admin Page
+      </a>
```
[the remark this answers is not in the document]
jorge-sepulveda: added it to the routerLink

libs/geoservices/covid/src/lib/pages/maps/components/lockdown-map/lockdown-map.component.ts
jorge-sepulveda: fixed

libs/geoservices/covid/src/lib/pages/maps/components/lockdown-map/lockdown-map.component.ts
jorge-sepulveda: fixed

libs/geoservices/covid/src/lib/pages/maps/components/lockdown-map/lockdown-map.component.ts
jorge-sepulveda: I noticed that I'm not using this function at the moment. I will remove it for now. If I need to re add it I will implement it using LockdownService

libs/geoservices/covid/src/lib/pages/maps/components/lockdown-map/lockdown-map.component.ts
jorge-sepulveda: the indices are string values(fips codes). I don't see how to make use of entries without using a for loop.

libs/geoservices/covid/src/lib/pages/maps/components/lockdown-map/lockdown-map.component.ts
jorge-sepulveda: got it chained

libs/geoservices/covid/src/lib/pages/maps/components/lockdown-map/lockdown-map.component.ts
Proantagonist: Seems like you maybe figured this out, but I'll add on this:

`Object.entries` will return an array of both the key and referenced object.

```js
this.statData = Object.entries(this.unformattedData)
  .map(([key, entry]) => {
    return entry.lockdownInfo
           .filter((item) => item !== null)
           .map((item) => {
              return {
                 updated: item.updated,
                 created: item.created,
                 guid: item.guid
            }
         })
   })
```

libs/geoservices/covid/src/lib/pages/maps/components/lockdown-map/lockdown-map.component.ts
jorge-sepulveda: removed bracket notation and used dot notation wherever applicable on the object.

libs/geoservices/covid/src/lib/pages/maps/components/lockdown-map/lockdown-map.component.ts
jorge-sepulveda: moved lockdownrecords into the map block

libs/geoservices/covid/src/lib/pages/maps/components/lockdown-map/lockdown-map.component.ts
jorge-sepulveda: set these as County Type objects

libs/geoservices/covid/src/lib/pages/maps/components/lockdown-map/lockdown-map.component.ts
```diff
@@ -77,52 +78,41 @@ export class LockdownMapComponent implements OnInit {
...
-      return lockdownData;
-    });
-  }
-
   public getStats(): void {
     const statsUrl = 'https://nodes.geoservices.tamu.edu/api/covid/counties/stats';
-    const requests = forkJoin([this.http.get(statsUrl)]);
-    requests.pipe(take(1)).subscribe(([recievedData]) => {
+    const requests = forkJoin([this.http.get<DataRecord[]>(statsUrl)]);
+    requests.subscribe(([recievedData]) => {
```
Proantagonist: Like the lockdowns request you were making previously which was removed due to it not being used, this URL should not be composed in the component. `CountiesService` from `@tamu-gisc/geoservices/data-access` exposes the `getCountyStats` method which returns the same data.

This would possibly eliminate the need to inject the `HttpClient` from this component. 

Additionally, `forkJoin` is not necessary as you're only making a single subscription. The http request can be subscribed to directly.

```js

this.http.get('url').subscribe((response) => {
   // Result here
});
```

Referenced review comment:

https://github.com/TamuGeoInnovation/Tamu.GeoInnovation.js.monorepo/pull/52#discussion_r442996485

jorge-sepulveda: resolved

libs/geoservices/covid/src/lib/pages/maps/components/lockdown-map/lockdown-map.component.ts
```diff
@@ -110,9 +110,8 @@ export class LockdownMapComponent implements OnInit {
   }
 
   public getStats(): void {
-    const statsUrl = 'https://nodes.geoservices.tamu.edu/api/covid/counties/stats';
-    const requests = forkJoin([this.http.get<DataRecord[]>(statsUrl)]);
-    requests.subscribe(([recievedData]) => {
+    const statService = this.countyService.getCountyStats();
+    const statSubscription = statService.subscribe((recievedData) => {
```
Proantagonist, line 114: Keeping the subscription to httpclient's observables is not necessary since they are self-completing.

Storing subscription references would be of some use if you'd want to manually unsubscribe from long-lived subscriptions at some point during or at the end of the component's life cycle. However, using the `takeUntil` operator to terminate the subscription is the preferred method.

You also  don't need to store the reference to the observable either, unless you plan to do things with it down the line, but since we're only interested in the value, that's not the case here.

So it should look like:

```js
this.countyService.getCountyStats().subscribe((receivedData) => {
   // Do stuff here.
})
```

jorge-sepulveda: resolved

PR #97, opened 2021-02-10
feat(ues-angular): Added Aggiemap layers

Proantagonist:
Added Aggiemap layers to the UES map under the category of "Aggiemap". These layers include:

- Construction Zone
- Points of Interest
- Restrooms
- Lactation Rooms
- Visitor Parking
- Accessible Entrances
- Emergency Phones
- Physical Distance Study Area

Finally consolidated the dev and prod environments under definitions so they're shared on both development serve and production build. No more forgetting to add it in two places.


atharmon (approved):
Everything looks good; nothing stands out

PR #116, opened 2021-05-19
ci(ues): Continuous integration implementation for the UES applications. 

IsaiahV98:
What: Implements a continuous integration solution via the usage of NX tools, and Azure Pipelines. 
Why:  Allows for the building of the UES application off of recent pr changes, and can be adapted for other applications. 
How: Utilizing the NX affected subset of commands: Built applications, and added build tags to azure CI pipelines.


Proantagonist (changes requested):
Run prettier on the affected files. Some minor EOL (end of line) issues.

Proantagonist (approved):

Proantagonist:
👍 👍 👍 👍

PR #123, opened 2021-05-22
maint(oidc): Scope cleanup for the OIDC apps/libraries

Proantagonist:
Several OIDC applications and libraries have now been superseded by NestJS variants or are no longer necessary. The full list is as follows:

- `oidc-provider-angular` (app). Not to be confused with `oidc-admin-angular` (app) which remains untouched.
- `oidc-provider-express` (app). This was the first iteration of the IDP and has been supeceded by `oidc-provider-nest`.
- `oidc-provider` (lib). Not to be confused with `oidc-povider-nestjs` (lib) which remains untouched.


atharmon (approved):
These old implementations of the IdP may be removed.

PR #128, opened 2021-05-27
 ci(ues): Edited ENV files for UES applications

IsaiahV98:
Updated the Env files to reference the correct API URL


Proantagonist (approved):
🚀 🚀 🚀 🚀 🚀 🚀

PR #131, opened 2021-06-03
feat(oidc-provider-nest): NodeMailer can now use Gmail to send emails

atharmon:
Requires a `mailerconfig.ts` file in the environments directory with an exported member like:
`export const mailerConfig = {
  user: string (email you've setup Google API for,
  accessToken: string,
  clientId: string,
  clientSecret: string,
  refreshToken: string
};`
You'll need to go through some steps to setup `mailerConfig` for your own email or I can show you it in action with mine.

### DIFF
--- a/apps/oidc-provider-nest/src/main.ts
+++ b/apps/oidc-provider-nest/src/main.ts
@@ -77,6 +77,13 @@ async function bootstrap() {
       type: 'string',
       nargs: 1
     })
+    .option('m', {
+      alias: 'email-provider',
+      description: 'The email provider (ethereal / tamu / gmail)',
+      type: 'string',
+      demandOption: true,
+      choices: ['ethereal', 'tamu', 'google']
+    })
     .help()
     .alias('help', 'h');
 
@@ -106,7 +113,9 @@ async function bootstrap() {
   const dir = join(__dirname, 'assets/views');
 
   // This will setup the Mailer (gmail or ethereal)
-  Mailer.build('gmail', mailerConfig);
+  // Mailer.build('gmail', mailerConfig);
+  Mailer.build('ethereal');
+  Mailer.sendPasswordResetConfirmationEmail('atharmon@tamu.edu');
 
   // This will set the default time step for otplib to 5 minutes
   TwoFactorAuthUtils.build();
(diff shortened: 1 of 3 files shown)


apps/oidc-provider-nest/src/main.ts
```diff
@@ -103,6 +112,11 @@ async function bootstrap() {
 
   const dir = join(__dirname, 'assets/views');
 
+  // This will setup the Mailer (gmail or ethereal)
+  // Mailer.build('gmail', mailerConfig);
+  Mailer.build('ethereal');
+  Mailer.sendPasswordResetConfirmationEmail('atharmon@tamu.edu');
```
Proantagonist, line 118: What's up with this hard-coded email?

libs/oidc/common/src/lib/utils/email/mailer.util.ts
```diff
@@ -3,18 +3,59 @@ import Mail from 'nodemailer/lib/mailer';
...
+      case 'ethereal':
+        // use ethereal email and print out links to console (debugging / dev)
+        Mailer.transporter = nodemailer.createTransport({
+          host: 'smtp.ethereal.email',
+          port: 587,
+          secure: false,
+          auth: {
+            user: 'kaitlyn.schimmel@ethereal.email',
+            pass: 'Pe6D9DhkgUDyqyBMeg'
+          }
```
Proantagonist, line 35: I can only assume this shouldn't be here?

Proantagonist (changes requested):
Gotta fix some hard-coded configs

libs/oidc/common/src/lib/utils/email/mailer.util.ts
atharmon: That depends. It's a free account that was automatically generated. Should it be there? No. Does it really matter? Not particularly.

apps/oidc-provider-nest/src/main.ts
atharmon: Was probably testing the functionality of the different mailing options

libs/oidc/common/src/lib/utils/email/mailer.util.ts
Proantagonist: image

Proantagonist (approved):

PR #135, opened 2021-06-17
fix(oidc-client): Fixes for access token exchange

atharmon:
Created a middleware (TokenExchangeMiddleware) that handles checking for expired access and refresh tokens. If an expired access token is found it will check and see if the refresh token is still valid. If it is still valid, it will exchange it for a new access token and update req.user with the new TokenSet.


libs/oidc/client/src/lib/guards/roles.guard.ts
```diff
@@ -1,5 +1,6 @@
-import { CanActivate, ExecutionContext, Injectable } from '@nestjs/common';
+import { CanActivate, ExecutionContext, Injectable, UnauthorizedException } from '@nestjs/common';
 
+import { AuthUtils } from '../../../../common/src/lib/utils/auth/auth.util';
```
Proantagonist, line 3: Cannot use relative lib imports.

This should show up as a warning in VSC.

Rule: `nx-enforce-module-boundaries`

libs/oidc/client/src/lib/middleware/token-exchange.middleware.ts
```diff
@@ -0,0 +1,29 @@
+import { Injectable, NestMiddleware } from '@nestjs/common';
+import { Response, NextFunction } from 'express';
+
+import { OpenIdClient } from '../auth/open-id-client';
+import { AuthUtils } from '../../../../common/src/lib/utils/auth/auth.util';
```
Proantagonist, line 5: Cannot import libs using relative paths:

Rule: `nx-enforce-module-boundaries`

libs/oidc/client/src/lib/middleware/token-exchange.middleware.ts
```diff
@@ -0,0 +1,29 @@
...
+
+@Injectable()
+export class TokenExchangeMiddleware implements NestMiddleware {
+  public async use(request: any, response: Response, next: NextFunction) {
```
Proantagonist, line 9: No use of `any`

Rule: `no-any`

libs/oidc/client/src/lib/middleware/token-exchange.middleware.ts
```diff
@@ -0,0 +1,29 @@
...
+    if (request.user) {
+      const tokenIntrospectionResult = await OpenIdClient.client.introspect(request.user.access_token);
+      if (tokenIntrospectionResult.active == false) {
```
Proantagonist, line 12: Should use triple equals.

libs/oidc/common/src/lib/utils/auth/auth.util.ts
```diff
@@ -0,0 +1,14 @@
+import { TokenSet } from 'openid-client';
+
+export class AuthUtils {
+  public static updateTokenSet(request: any, tokenSet: TokenSet) {
```
Proantagonist, line 4: No use of `any`

Rule: `no-any`

Proantagonist (changes requested):

libs/oidc/client/src/lib/guards/roles.guard.ts
atharmon: We need to go over how to get my linter working properly

Proantagonist (approved):
👍

PR #137, opened 2021-06-29
feat(oidc-provider / oidc-client): Support for back channel logout

atharmon:
Enabled back channel logout support. 

Required adding some additional logic to the yargs initialization system as well as some additions to the mega client-metadata.service class. 

Added an additional route to the oidc-client lib which will logout the user.

### DIFF
--- a/libs/oidc/common/src/lib/services/client-metadata/client-metadata.service.ts
+++ b/libs/oidc/common/src/lib/services/client-metadata/client-metadata.service.ts
@@ -14,7 +14,9 @@ import {
   ResponseType,
   ResponseTypeRepo,
   TokenEndpointAuthMethod,
-  TokenEndpointAuthMethodRepo
+  TokenEndpointAuthMethodRepo,
+  BackchannelLogoutUri,
+  BackchannelLogoutUriRepo
 } from '../../entities/all.entity';
 
 @Injectable()
@@ -24,7 +26,8 @@ export class ClientMetadataService {
     private readonly grantTypeRepo: GrantTypeRepo,
     private readonly redirectUriRepo: RedirectUriRepo,
     private readonly responseTypeRepo: ResponseTypeRepo,
-    private readonly tokenEndpointRepo: TokenEndpointAuthMethodRepo
+    private readonly tokenEndpointRepo: TokenEndpointAuthMethodRepo,
+    private readonly backchannelUriRepo: BackchannelLogoutUriRepo
   ) {}
 
   // ClientMetadata functions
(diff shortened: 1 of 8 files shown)


libs/oidc/common/src/lib/services/client-metadata/client-metadata.service.ts
```diff
@@ -48,27 +51,35 @@ export class ClientMetadataService {
...
       responseTypes: ['code'],
-      token_endpoint_auth_method: 'client_secret_basic'
+      token_endpoint_auth_method: 'client_secret_basic',
+      backchannelLogoutUris: [backchannelLogoutUri]
     };
     const grants = await this.findGrantTypeEntities(adminMetadata.grants);
     const redirectUris = await this.createRedirectUriEntities(adminMetadata.redirectUris);
     const responseTypes = await this.findResponseTypeEntities(adminMetadata.responseTypes);
     const token_endpoint_auth_method = await this.findTokenEndpointAuthMethod(adminMetadata.token_endpoint_auth_method);
+    const backchannelUris = await this.createBackchannelLogoutUri(adminMetadata.backchannelLogoutUris);
```
Proantagonist, line 73: `createBackchannelLogoutUri` does not have any async behavior.

libs/oidc/common/src/lib/services/client-metadata/client-metadata.service.ts
```diff
@@ -236,6 +256,21 @@ export class ClientMetadataService {
...
+      const backchannelLogoutUri: Partial<BackchannelLogoutUri> = {
+        url: value
+      };
+      const logoutUri = this.backchannelUriRepo.create(backchannelLogoutUri);
+      backchannelLogoutUris.push(logoutUri);
+    });
+
+    return backchannelLogoutUris;
+  }
+
```
Proantagonist, line 273: Nothing about this block is async. Can cause a race condition since 271 will not necessary return after the  `_backchannelLogoutUris` are completely mapped.

Wrong use of `Array.map()`.

libs/oidc/common/src/lib/services/client-metadata/client-metadata.service.ts
```diff
@@ -81,12 +92,16 @@ export class ClientMetadataService {
       redirectUri.save();
     });
 
+    clientMetadata.backchannelLogoutUris.map((backchannelUri) => {
+      backchannelUri.save();
+    });
+
```
Proantagonist, line 98: `backchannelUri.save()` is async and line 99 *can* execute before the save is complete. Potential race condition and the save on line 99 can fail since the relationships will yet exist.

Use `await Promise.all(promisesArray);` to ensure those relationships are inserted before proceeding. This applies to inserting client metadata above this block.

Proantagonist (changes requested):

Proantagonist (approved):
👍

PR #142, opened 2021-07-20
ci: Add check for verifying PM2 installed as a service. 

IsaiahV98:
Added a check in the pm2_process.ps1 file that checks for whether or not PM2 exists as a windows service, and if not throws a more informative error.


Proantagonist (approved):
LGTM

PR #143, opened 2021-07-20
 ci: PM2 launch on windows start-up via NSSM, as well as restart if crash. 

IsaiahV98:
Created a startup.batch file in the monorepo which contains the start-up procedure for NSSM. Modified said procedure to query if PM2 was already running, and if so don't attempt to resurrect. This was done to mitigate an issue with the resurrect command in which it was stuck in a infinite loop crashing/slowing down the server.

### DIFF
--- a/libs/assets/batch/startup.bat
+++ b/libs/assets/batch/startup.bat
@@ -1,0 +1,9 @@
+sc query "PM2" | find "RUNNING"
+
+if "%ERRORLEVEL%"=="0" (
+    echo Already Running
+    pause
+) else (
+    pm2 resurrect
+    pause
+)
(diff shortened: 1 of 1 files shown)


libs/assets/batch/startup.bat
```diff
@@ -0,0 +1,9 @@
+sc query "PM2" | find "RUNNING"
```
Proantagonist, line 1: This will only work IF this script is running. 

The windows service process named "PM2" created by NSSM  is *this* script. It will not check if the actual pm2 process is running or not.

Proantagonist (changes requested):

libs/assets/batch/startup.bat
```diff
@@ -0,0 +1,8 @@
+pm2 status | find /i "Current process list running is not in sync with saved list. App effluent-api dispatch-api recycling-api operations-api differs. Type 'pm2 save' to synchronize."
```
Proantagonist, line 1: Process list is not the same for all hosts

Proantagonist (changes requested):

libs/assets/batch/startup.bat
IsaiahV98: Removed host specific process list refrence.

libs/assets/batch/startup.bat
IsaiahV98: Addressed per discussion

PR #160, opened 2021-10-19
chore(deps): New package lock file

Proantagonist:
The old was installing a weird version of Nx that was prohibiting projects builds. This one also fixes various additional vulnerability that NPM warns about.


Proantagonist:
Closed instead of merging

PR #185, opened 2022-02-05
testing(CPA): Unit tests added for CPA lib data-access services. 

IsaiahV98:
The following PR was for getting back into the groove of testing. Establishing how to test services with dependencies on other services and Http Client. 

### **Process for testing a service dependent on another service:** 
1. Create a skeleton mock. 
`const spyForDependentService= { someFunction: jest.fn() }`
 2. In the beforeEach establish the providers array:
```
beforeEach(() => {
  TestBed.configureTestingModule({
    providers: [
      ServiceToBeTested,
        {
          provide: DependentService,
          useValue: spyForDependentService
        }
      ]
    });
});
```
3. If the logic of the service has different actions based on a value condition of the dependent service, don't inject the service being tested in the beforeEach configureTestingModule statement. Instead inject it in the individual test cases after mocking the return value for said dependent services function call. 

### **Process for testing a HttpClient request (based on https://angular-training-guide.rangle.io/testing/services/http/using-httptestingmodule):** 
1. Import HttpClientTestingModule and HttpTestingController
` import { HttpClientTestingModule, HttpTestingController } from '@angular/common/http/testing';`
2. Declare HttpTestingController in the outer describe statement, import HttpClientTestingModule in the configureTestingModule statement, and inject the HttpTestingController into the testbed. 
```
beforeEach(() => {
  TestBed.configureTestingModule({
    imports: [HttpClientTestingModule],
    providers: [
      ServiceToBeTested,
    ]
  });
  httpTestingController = TestBed.inject(HttpTestingController);
});
```
3. Make a dummy response and value, call the function in the service to be tested that makes a HttpClient call, subscribe to it, listen for specific response. 
```
const dummyValue = whatever value needed for function call;
const dummyResponse = [
  {
    whateverneeded: 'Foo',
  }
];

service.functionbeingtested(dummyValue).subscribe((interface/object) => {
  expect(interface/object.whateverneeded).toBe('Foo');
});
```
4. Call the httpTestingController with the expected API url the response will be made on, test the http method, and flush the response. 
```
const req = httpTestingController.expectOne(APIURL);
expect(req.request.method).toEqual('GET/POST/Whatever');
req.flush(dummyResponse);
```

### DIFF
--- a/libs/cpa/ngx/data-access/src/lib/services/participant/participants.service.spec.ts
+++ b/libs/cpa/ngx/data-access/src/lib/services/participant/participants.service.spec.ts
@@ -47,8 +47,7 @@ describe('ParticipantService', () => {
       expect(participants[0].name).toBe('Foo');
     });
 
-    const req = httpTestingController.expectOne(`${service.resource}/workshop`);
-    expect(req.request.method).toEqual('POST');
+    const req = httpTestingController.expectOne({ method: 'POST', url: `${service.resource}/workshop` });
     req.flush(dummyResponse);
   });
 
@@ -69,8 +68,7 @@ describe('ParticipantService', () => {
       expect(participants.length).toBe(1);
     });
 
-    const req = httpTestingController.expectOne(`${service.resource}/workshop/${dummyStringValue}`);
-    expect(req.request.method).toEqual('GET');
+    const req = httpTestingController.expectOne({ method: 'GET', url: `${service.resource}/workshop/${dummyStringValue}` });
     req.flush(dummyResponse);
   });
 });
--- a/libs/cpa/ngx/data-access/src/lib/services/participant/participants.service.spec.ts
+++ b/libs/cpa/ngx/data-access/src/lib/services/participant/participants.service.spec.ts
@@ -1,16 +1,75 @@
-import { TestBed } from '@angular/core/testing';
+import { fakeAsync, TestBed } from '@angular/core/testing';
+import { HttpClientTestingModule, HttpTestingController } from '@angular/common/http/testing';
 
+import { EnvironmentService } from '@tamu-gisc/common/ngx/environment';
 import { ParticipantService } from './participants.service';
+
+const spyEnvironment = {
+  value: jest.fn()
+};
 
 describe('ParticipantService', () => {
   let service: ParticipantService;
+  let httpTestingController: HttpTestingController;
 
   beforeEach(() => {
-    TestBed.configureTestingModule({});
-    service = TestBed.inject(ParticipantService);
+    TestBed.configureTestingModule({
+      imports: [HttpClientTestingModule],
+      providers: [
+        ParticipantService,
+        {
+          provide: EnvironmentService,
+          useValue: spyEnvironment
+        }
+      ]
+    });
+    httpTestingController = TestBed.inject(HttpTestingController);
   });
 
-  it('should be created', () => {
-    expect(service).toBeTruthy();
+  afterEach(() => {
+    httpTestingController.verify();
   });
+
+  it('createParticipantForWorkshop()', fakeAsync(() => {
+    const dummyStringValue = 'Foo';
+    const dummyResponse = [
+      {
+        name: 'Foo',
+        workshops: []
+      }
+    ];
+
+    spyEnvironment.value.mockReturnValue(dummyStringValue);
+    service = TestBed.inject(ParticipantService);
+
+    service.createParticipantForWorkshop(dummyStringValue, dummyStringValue).subscribe((participants) => {
+      expect(participants[0].name).toBe('Foo');
+    });
+
+    const req = httpTestingController.expectOne(`${service.resource}/workshop`, 'reee');
+    expect(req.request.method).toEqual('POST');
+    req.flush(dummyResponse);
+  }));
+
+  it('getParticipantsForWorkshop()', fakeAsync(() => {
+    const dummyStringValue = 'Foo';
+    const dummyResponse = [
+      {
+        name: 'Foo',
+        workshops: []
+      }
+    ];
+
+    spyEnvironment.value.mockReturnValue(dummyStringValue);
+    service = TestBed.inject(ParticipantService);
+
+    service.getParticipantsForWorkshop(dummyStringValue).subscribe((participants) => {
+      expect(participants[0].name).toBe('Foo');
+      expect(participants.length).toBe(1);
+    });
+
+    const req = httpTestingController.expectOne(`${service.resource}/workshop/${dummyStringValue}`);
+    expect(req.request.method).toEqual('GET');
+    req.flush(dummyResponse);
+  }));
 });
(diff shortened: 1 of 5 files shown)


libs/cpa/ngx/data-access/src/lib/services/response/response.service.spec.ts
```diff
@@ -1,25 +1,53 @@
...
+  });
+
+  it('createResponse()', () => {
+    const dummyStringValue = 'Foo';
+    const dummyResponse = [
+      {
+        name: 'Foo',
+        workshops: []
+      }
+    ];
```
Proantagonist, line 40: Incorrect input and return type. I realize the service method being tested doesn't have a very useful implicit return type, but you are encouraged to look at the implementation and determine what kind of response comes back, if any, and have more accurate tests.

libs/cpa/ngx/data-access/src/lib/services/scenario/scenario.service.spec.ts
```diff
@@ -1,25 +1,52 @@
...
+        workshops: []
+      }
+    ];
+
+    spyEnvironment.value.mockReturnValue(dummyStringValue);
+    service = TestBed.inject(ScenarioService);
+
+    service.create({}).subscribe((response) => {
+      expect(response[0].name).toBe('Foo');
+    });
```
Proantagonist, line 46: Incorrect return type and incorrect assertion

libs/cpa/ngx/data-access/src/lib/services/scenario/scenario.service.spec.ts
```diff
@@ -1,25 +1,52 @@
...
 
-  it('should be created', () => {
-    const service: ScenarioService = TestBed.get(ScenarioService);
-    expect(service).toBeTruthy();
+    const req = httpTestingController.expectOne(`${'Fooscenarios'}`);
```
Proantagonist, line 48: This is definitely not the endpoint

libs/cpa/ngx/data-access/src/lib/services/snapshot/snapshot.service.spec.ts
```diff
@@ -1,25 +1,53 @@
...
+        workshops: []
+      }
+    ];
+
+    spyEnvironment.value.mockReturnValue(dummyStringValue);
+    service = TestBed.inject(SnapshotService);
+
+    service.create({}).subscribe((response) => {
+      expect(response[0].name).toBe('Foo');
+    });
```
Proantagonist, line 47: Invalid return type and assertion

libs/cpa/ngx/data-access/src/lib/services/workshop/workshop.service.spec.ts
```diff
@@ -1,25 +1,53 @@
...
+        workshops: []
+      }
+    ];
+
+    spyEnvironment.value.mockReturnValue(dummyStringValue);
+    service = TestBed.inject(WorkshopService);
+
+    service.createWorkshop({ guid: 'bruh' }).subscribe((response) => {
+      expect(response[0].name).toBe('Foo');
+    });
```
Proantagonist, line 47: Incorrect return type and assertion

Proantagonist:
Discussion:

If we can write tests with any random input values, endpoint urls, etc, and mock any kind of response, and make assertions that effectively are `true == true`, then what is the value of these tests. I have a really hard time understanding why we should dedicate any amount of time to these tests except for maybe coverage reporting.

Proantagonist (changes requested):

libs/cpa/ngx/data-access/src/lib/services/participant/participants.service.spec.ts
```diff
@@ -1,16 +1,75 @@
 import { TestBed } from '@angular/core/testing';
+import { HttpClientTestingModule, HttpTestingController } from '@angular/common/http/testing';
 
+import { EnvironmentService } from '@tamu-gisc/common/ngx/environment';
 import { ParticipantService } from './participants.service';
 
+const spyEnvironment = {
```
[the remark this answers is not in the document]
IsaiahV98: Jest as a testing library makes no real distinction between spys and mocks (both result in mock functions): https://jestjs.io/docs/jest-object#jestspyonobject-methodname

IsaiahV98:
> Discussion:
> 
> If we can write tests with any random input values, endpoint urls, etc, and mock any kind of response, and make assertions that effectively are `true == true`, then what is the value of these tests. I have a really hard time understanding why we should dedicate any amount of time to these tests except for maybe coverage reporting.

The value in these tests came from me attempting to figure how to deal with angular service dependencies utilizing Jest. Identifying and documenting the correct way deal with scenarios in which a service is dependent on another service, and identifying and documenting the correct way to deal with scenarios in which a service is dependent on Angular’s  HttpClient via the utilization of the HttpClientTestingModule and HttpTestingController. The files and corresponding functions under test here have their logic completely dependent on dependencies, which means to my understanding that there is basically nothing to test for in the scope of a unit test. So, for future testing there is no point outside of increasing testing coverage for unit testing these kinds of files/functions. 

However, now that it is documented how to properly utilize HttpClientTestingModule and HttpTestingController, we can in the future refer back to this documentation and code example for scenarios in which logic is not completely dependent on dependencies. Such as pipe calls on HttpClient requests and such.

libs/cpa/ngx/data-access/src/lib/services/participant/participants.service.spec.ts
```diff
@@ -1,16 +1,75 @@
...
+  });
+
+  it('createParticipantForWorkshop()', () => {
+    const dummyStringValue = 'Foo';
+    const dummyResponse = [
+      {
+        name: 'Foo',
+        workshops: []
+      }
+    ];
```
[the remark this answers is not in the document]
IsaiahV98: > This is also where my understanding of the purpose of these tests falls apart. The dummyResponse is the incorrect return type for the service call createParticipantForWorkshop. The service method returns an object representing a participant, not an array of participants which is what you're testing for.

This is something I should have payed closer attention to, and will work on fixing.

libs/cpa/ngx/data-access/src/lib/services/participant/participants.service.spec.ts
```diff
@@ -1,16 +1,75 @@
...
+      ]
+    });
+    httpTestingController = TestBed.inject(HttpTestingController);
+  });
+
+  afterEach(() => {
+    httpTestingController.verify();
+  });
+
+  it('createParticipantForWorkshop()', () => {
```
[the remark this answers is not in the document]
IsaiahV98: Fixed

libs/cpa/ngx/data-access/src/lib/services/participant/participants.service.spec.ts
```diff
@@ -1,16 +1,75 @@
...
+
+    spyEnvironment.value.mockReturnValue(dummyStringValue);
     service = TestBed.inject(ParticipantService);
+
+    service.createParticipantForWorkshop(dummyStringValue, dummyStringValue).subscribe((participants) => {
+      expect(participants[0].name).toBe('Foo');
+    });
+
+    const req = httpTestingController.expectOne(`${service.resource}/workshop`);
+    expect(req.request.method).toEqual('POST');
```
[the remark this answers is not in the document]
IsaiahV98: Works, and applied for all tests.

PR #207, opened 2022-04-20
feat(oidc-provider-nest): JWT-based authentication

atharmon:
IdP now uses JWT-based authentication. Several features (password reset, 2FA, secret questions, various logging) have been disabled for the time being so we can test the IdP.


libs/oidc/provider/src/lib/services/common/common.service.ts
```diff
@@ -1,8 +1,9 @@
-import { getConnection, EntitySchema, FindConditions, BaseEntity } from 'typeorm';
+import { getConnection, EntitySchema, FindConditions } from 'typeorm';
 
-export type TypeORMEntities = string | Function | (new () => unknown) | EntitySchema<unknown>;
+export type TypeORMEntities = string | (new () => unknown) | EntitySchema<unknown>;
 
 export abstract class CommonService {
+  // TODO: Is this even used anymore? Everything looks very shitty - Aaron H (4/21/22)
```
Proantagonist, line 6: `Ctrl + Shift + F12` on member name. Doesn't appear to be used anywhere, so probably not used anymore?

libs/oidc/common/src/lib/services/account/account.service.ts
```diff
@@ -19,3 +20,20 @@ export class StaticAccountService extends CommonService {
...
   }
 }
+
+@Injectable()
+export class AccountService {
+  constructor(public readonly accountRepo: AccountRepo, public readonly roleRepo: RoleRepo) {}
+
+  public default() {
+    return 'AccountService';
+  }
```
Proantagonist, line 30: ???

libs/oidc/common/src/lib/services/user/user.service.ts
```diff
@@ -30,8 +30,8 @@ import * as deepmerge from 'deepmerge';
 
 @Injectable()
 export class UserService {
-  private IPSTACK_APIKEY = '1e599a1240ca8f99f0b0d81a08324dbb';
-  private IPSTACK_URL = 'http://api.ipstack.com/';
+  // private IPSTACK_APIKEY = '1e599a1240ca8f99f0b0d81a08324dbb';
+  // private IPSTACK_URL = 'http://api.ipstack.com/';
```
Proantagonist, line 34: No secret keys of any kind in version control

libs/oidc/common/src/lib/services/user/user.service.ts
```diff
@@ -182,15 +182,16 @@ export class UserService {
     const resetRequest = await this.passwordResetRepo.create(_resetRequest);
     resetRequest.setToken();
 
-    // TODO: Is this subscribe a potential source of problems later on? Should we have to unsubscribe?
-    this.httpService.get(`${this.IPSTACK_URL}${ip}?access_key=${this.IPSTACK_APIKEY}`).subscribe((observer) => {
-      const location = observer.data.country_name;
+    // TODO: Disabled for now because of too many API calls to ipstack; maybe we investigate other providers or scrap this functionality
+    // Is this subscribe a potential source of problems later on? Should we have to unsubscribe?
+    // this.httpService.get(`${this.IPSTACK_URL}${ip}?access_key=${this.IPSTACK_APIKEY}`).subscribe((observer) => {
+    //   const location = observer.data.country_name;
```
Proantagonist, line 188: It shouldn't be.

 `.get()` should be a unicast observable which means it should not keep emitting values past the initial time of execution.

Only way many API requests are going out is if the `sendPasswordResetEmail` method is called many times.

libs/oidc/common/src/lib/utils/email/mailer.util.ts
```diff
@@ -1,12 +1,12 @@
...
 import Mail from 'nodemailer/lib/mailer';
 
-import { User, UserPasswordReset } from '@tamu-gisc/oidc/common';
+import { User, UserPasswordReset } from '../../entities/all.entity';
 import { IMailroomEmailOutbound } from '@tamu-gisc/mailroom/common';
 
 export type NodeMailerServices = 'ethereal' | 'gmail' | 'tamu-relay';
 
-const fs = require('fs');
+// const fs = require('fs'); TODO: disabled for now, what was this here for anyway?
```
Proantagonist, line 9: Just get rid of it then. Not used anywhere.

libs/oidc/common/src/lib/utils/web/jwt.util.ts
```diff
@@ -34,6 +34,6 @@ interface IDecoded {
   aud: string;
   iat: string;
   jti: string;
-  events: {};
+  events: object;
```
Proantagonist, line 37: Might causes issues downstream since it doesn't allow property value re-assignment nor can you add properties that the OG object does not have.

```js

interface ValidObject {
  [key: string] : boolean | string | number | ValidObject;
}

```

libs/oidc/provider/src/lib/adapters/oidc.adapter.ts
```diff
@@ -163,26 +146,36 @@ export class OidcAdapter {
...
+interface IOidcAdapter {
+  upsert(id, payload, expiresIn);
+  find(id);
+  findByUserCode(userCode);
+  findByUid(uid);
+  destroy(id);
+  revokeByGrantId(grantId);
+  consume(id);
+}
+
```
Proantagonist, line 178: Nothing else implements this interface other than the single `OidcAdapter`. Are there cases where this will be implemented elsewhere?

libs/oidc/provider/src/lib/adapters/oidc.adapter.ts
```diff
@@ -129,30 +120,22 @@ export class OidcAdapter {
...
-      };
-    } else {
-      return {
-        ...(found.consumedAt ? { consumed: true } : undefined)
-      };
-    }
+    return {
+      ...JSON.parse(found.data),
+      ...(found.consumedAt ? { consumed: true } : undefined)
+    }; // is there ever a time in which data is null? If so how do we return?
```
Proantagonist, line 134: I don't understand the "how do we return" part.

libs/oidc/provider/src/lib/controllers/access-token/access-token.controller.ts
```diff
@@ -1,13 +1,16 @@
...
-    const grantId = params.grantId;
+    // TODO: Reimplement a means to revoke access. With JWT this would be adding a JWT access token to a black list
+    // const grantId = params.grantId;
+    // OpenIdProvider.provider.AccessToken.revokeByGrantId(grantId);
 
-    OpenIdProvider.provider.AccessToken.revokeByGrantId(grantId);
+    return {
+      status: 200,
+      message: 'Not implemented at the moment'
+    };
```
Proantagonist, line 14: Prefer using built in exceptions. 

In this case, `NotImplementedException` which returns a standard message and appropriate status code.

libs/oidc/provider/src/lib/controllers/interaction/interaction.controller.ts
```diff
@@ -237,8 +205,58 @@ export class InteractionController {
...
+
+      if (details.missingOIDCScope) {
+        grant.addOIDCScope((details.missingOIDCScope as string[]).join(' '));
+      }
+      if (details.missingOIDCClaims) {
+        grant.addOIDCClaims(details.missingOIDCClaims);
+      }
+      if (details.missingResourceScopes) {
+        // eslint-disable-next-line no-restricted-syntax
+        // TODO: Enable this again
```
Proantagonist, line 241: ESLint doesn't give me any issues. Might not actually be needed?

Can't tell what about that is supposed to be restricted syntax

Proantagonist (changes requested):

libs/oidc/common/src/lib/services/user/user.service.ts
atharmon: I don't know where it's coming from but I do get emails from ipstack saying I'm using almost all of my free credits every month: "Please be aware that you have reached at least 101% of your available API Request volume. ". Unless someone using my API key I don't know where that usage is coming from.

libs/oidc/provider/src/lib/adapters/oidc.adapter.ts
atharmon: I don't believe so. Just remove it?

libs/oidc/common/src/lib/services/user/user.service.ts
atharmon: The EnvironmentService would work great here, but seems to be limited to Angular projects only. Is there an existing EnvironmentService for Nest or should I consider making one?

Proantagonist: There is not one for Nest. I can't remember correctly but I *think* I tried implementing one some time ago and had an issue with injection tokens not being available outside the module where they were set which put a damper on being able to declare the environment on the root module and accessing it on some child module downstream. Not sure if anything's changed. 

But yes, either a service needs to be made OR just pass those values down as provider tokens. I don't think this PR should be dependent on an additional project so for now I say just go with the latter and we can figure out the environment service later. There's an example in the `veoride-data-api.module.ts` file.

Proantagonist (approved):

Issue #220, opened 2022-05-05
oidc-provider-nest: Fix client reset on deployment


atharmon:

Proantagonist:
Duplicate of #214

Issue #214, opened 2022-05-04
oidc-provider-nest: Clients are being removed on instance restart


Proantagonist:
On idp-dev, previously used clients are invalid on service restart. This could be due to clients being dropped on server initialization.

Registered users are unaffected.

PR #223, opened 2022-05-06
fix(oidc): Fixed data field getting set to NULL on deployment

atharmon:
- Updated all entities used by oidc-provider that had data properties to have a length of MAX instead of max.


Proantagonist (approved):

Proantagonist:
Addresses #214  and #220

PR #225, opened 2022-05-10
feat: Environment access library for NestJS apps/libs

Proantagonist:
This is the NestJS counterpart of the existing Angular environment service that allows registering environment variables in a single place and having them accessible through a service for apps/libs.


atharmon (approved):
:thumbsup:

PR #226, opened 2022-05-12
feat(oidc-common): JWT auth module

Proantagonist:
Implements a generic authentication module that can be applied to existing nestjs applications. Exposes a JWT auth guard that validates a request's bearer JWT against a public key endpoint. 

Documentation on the root of the library.

libs\oidc\common\README.md

### DIFF
--- a/libs/oidc/common/README.md
+++ b/libs/oidc/common/README.md
@@ -1,7 +1,53 @@
-# oidc-common
+# TAMU GISC OIDC Common
 
-This library was generated with [Nx](https://nx.dev).
+This library contains a series of utilities used in the OIDC provider NestJS application and some more generic authentication utilities for use in back-end data NestJS-based data API's .
 
-## Running unit tests
+## Library Scope
 
-Run `ng test oidc-common` to execute the unit tests via [Jest](https://jestjs.io).
+`@tamu-gisc/oidc/common`
+
+### Modules
+
+- AuthModule
+
+### Guards
+
+- JwtGuard
+
+## Setup
+
+In a root module of a NestJS application, import the `AuthModule` and provide the OIDC JWKS endpoint using the `forRoot` static method.
+
+```js
+import { Module } from '@nestjs/common';
+
+import { AuthModule } from '@tamu-gisc/common/oidc';
+
+import * as env from '../environments/environment';
+
+@Module({
+  imports: [AuthModule.forRoot({ jwksUrl: env.jwksUrl })]
+})
+export class AppModule {}
+```
+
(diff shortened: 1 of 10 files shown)


libs/oidc/common/README.md
```diff
@@ -1,7 +1,53 @@
...
+- JwtGuard
+
+## Setup
+
+In a root module of a NestJS application, import the `AuthModule` and provide the OIDC JWKS endpoint using the `forRoot` static method.
+
+```js
+import { Module } from '@nestjs/common';
+
+import { AuthModule } from '@tamu-gisc/common/oidc';
```
atharmon, line 24: @tamu-gisc/oidc/common not @tamu-gisc/common/oidc

libs/oidc/common/README.md
```diff
@@ -1,7 +1,53 @@
...
+  imports: [AuthModule.forRoot({ jwksUrl: env.jwksUrl })]
+})
+export class AppModule {}
+```
+
+This configuration takes care of initializing passport and registering a global JWT passport strategy that is implemented by the `JwtGuard`.
+
+## Utilization
+
+Apply the `JwtGuard` to any endpoints that should require a validated against the registered `AuthModule` public keys endpoint.
```
atharmon, line 38: A validated what?

libs/oidc/common/src/lib/modules/auth/auth.module.ts
```diff
@@ -0,0 +1,21 @@
+import { DynamicModule, Module } from '@nestjs/common';
+
+import { JwtStrategy } from '../../strategies/jwt/jwt.strategy';
+import { JWKS_URL } from '../tokens/jwks_url.token';
+
+@Module({})
+export class AuthModule {
+  public static forRoot(options: { jwksUrl: string }): DynamicModule {
```
atharmon, line 8: Is the return type here required? You don't usually include that stuff.

Proantagonist: It's not, but it makes it so the return object is typed whenever I am writing it out. `forRoot`, `forChild`, `register`, w/e you want to call initialization static methods can have whatever name, it just comes down to semantics. And since these methods can return anything and there isn't any interface `implements`, then the IDE can't know that it's supposed to return  (like OnInit that returns `void` and returning anything else violates the signature.)

libs/oidc/common/README.md
Proantagonist: Whoops. Fixed.

libs/oidc/common/README.md
Proantagonist: Reworded. Fixed.

atharmon (approved):

Issue #238, opened 2022-06-01
oidc-admin: Need a default admin user 


atharmon:
We need to include the creation of a default admin user for the `oidc-admin` projects so that we can log in and create new admins.

atharmon:
Fixed in PR #240.

PR #240, opened 2022-06-03
feat(oidc-provider-nest): Setup default values on startup

atharmon:
Addresses issue #238.

The `oidc-provider-nest` now uses values preset in `secrets.ts` to setup initial values for a default client (which would mostly always be the admin site) and a default admin user to access said default client.

### DIFF
--- a/apps/oidc-provider-nest/src/environments/environment.dev.ts
+++ b/apps/oidc-provider-nest/src/environments/environment.dev.ts
@@ -6,4 +6,4 @@ export const environment = {
 };
 
 export { devDbConfig as dbConfig } from './ormconfig';
-export { ADMIN_DEFAULTS as ADMIN_DEFAULTS } from './secrets';
+export { ADMIN_DEFAULTS } from './secrets';
--- a/libs/oidc/common/src/lib/services/user-role/user-role.service.ts
+++ b/libs/oidc/common/src/lib/services/user-role/user-role.service.ts
@@ -16,25 +16,46 @@ export class UserRoleService {
   ) {}
 
   public async insertDefaultRoles() {
-    const existing = await this.roleRepo.count();
+    // TODO: We prevent another role from being created outside of this method
+    // with a role equal to or greater than Admin level 99? - Aaron (6/6/22)
 
-    if (existing) {
+    // Look for an Admin role with this level
+    const existingAdmin = await this.roleRepo.findOne({
+      where: {
+        level: '99'
+      }
+    });
+
+    // If this level exists, do nothing
+    if (existingAdmin) {
       return;
+    } else {
+      // Role at this level not found, create it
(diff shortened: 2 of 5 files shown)


apps/oidc-provider-nest/src/environments/environment.dev.ts
```diff
@@ -6,3 +6,4 @@ export const environment = {
 };
 
 export { devDbConfig as dbConfig } from './ormconfig';
+export { ADMIN_DEFAULTS as ADMIN_DEFAULTS } from './secrets';
```
Proantagonist, line 9: Minor point: No need to alias the export with the same name.

libs/oidc/common/src/lib/services/user-role/user-role.service.ts
```diff
@@ -15,6 +15,94 @@ export class UserRoleService {
...
+
+      await this.clientRepo.save(client).then((val) => {
+        this.userRoleRepo
+          .create({
+            client: val,
+            user: defaultUser,
+            role: adminRole
+          })
+          .save();
+      });
```
Proantagonist, line 93: Implicitly or explicitly return promises in a `then` block, otherwise they will execute asynchronously.

Preferably, don't mix promises and async/await. Use one or the other.

libs/oidc/common/src/lib/services/user/user.service.ts
```diff
@@ -42,38 +42,68 @@ export class UserService {
...
+          questionText: 'In what town or city did you meet your spouse or partner?'
+        },
+        {
+          questionText: 'What is the middle name of your oldest child?'
+        }
+      ];
+
+      questions.forEach((question: Partial<SecretQuestion>) => {
+        this.questionRepo.create(question).save();
+      });
```
Proantagonist, line 105: This will run asynchronously. Create an array and `await Promise.allSettled()`

Proantagonist (changes requested):

apps/oidc-provider-nest/src/environments/environment.dev.ts
atharmon: Fixed.

libs/oidc/common/src/lib/services/user-role/user-role.service.ts
```diff
@@ -15,6 +15,94 @@ export class UserRoleService {
...
+        level: '99'
+      })
+      .save();
+
+    await this.roleRepo
+      .create({
+        name: 'Manager',
+        level: '80'
+      })
+      .save();
```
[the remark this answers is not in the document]
atharmon: Good catch. I think I've fixed it for now.

Proantagonist (approved):
🚀 🚀 🚀 🚀

PR #250, opened 2022-06-29
feat(maps-esri): Support OAuth authentication for ArcGIS secure resources

Proantagonist:
`LayerSources` now support a new `auth` property that is used by the esri map service to register their auth info's with the ArcGIS JS API `IdentityManager` which enables the API to implement the OAuth login flow for secure layer resources.

Implemented in `ues-operations-angular` and further removes the need to have "secure layers" in the ues-operations-nest API (#246 ). This PR should also 100% eliminate the need for `ues-operations-nest` app since it will no longer manage authentication using our  IdP but that will be removed in a future PR.


libs/common/types/src/lib/common-types.ts
```diff
@@ -231,6 +231,20 @@ interface PortalMapServerLayerSourceProperties {
...
+
+  /**
+   * Portal service base URL. The schema for the service will be pulled and all layers contained in the definition
+   * will be loaded.
+   *
+   * Example: https://service.domain/arcgis/rest/services/Folder/Folder/MapServer
+   */
+  url: string;
+
+  native?: any;
```
atharmon, line 249: Are `any` allowed?

Proantagonist: They should not. Fixed.

atharmon (approved):
👍

PR #256, opened 2022-07-06
fix(common-ngx-ui-notification): Notification item positioning

Proantagonist:
This bug was introduced in #114 when the ability to provide the general positioning for notification items (left, center, right) was added. Left and right positioning worked as it should, but center positioning was biasing towards the right. This PR fixes positioning for centered notifications. 

Before:

image

After:

image

Also took this opportunity to add documentation to the lib.


apps/aggiemap-angular/src/app/app.component.ts
```diff
@@ -1,13 +1,18 @@
...
   styleUrls: ['./app.component.scss']
 })
-export class AppComponent {
-  constructor(public analytics: Angulartics2GoogleAnalytics) {
+export class AppComponent implements OnInit {
+  constructor(public analytics: Angulartics2GoogleAnalytics, private ns: NotificationService) {
     analytics.startTracking();
   }
+
+  public ngOnInit(): void {
```
atharmon, line 15: I faintly remember you telling me long ago about having the return types defined on methods and how for something like `ngOnInit` that it's pointless to just remove it.

atharmon (approved):
👍

PR #257, opened 2022-07-06
fix(oidc-provider-nest): Added manual package.json for deployments

Proantagonist:
The auto-generated version did not have pinned versions which was causing an initialization error due to a breaking change in one of the dependencies.

This is a recurring issue with some deployments and the ultimate solution is coming up with a script that goes through the package-lock tree and picks out all pinned dependency versions or better yet, creates a `package-lock.json` from the root package-lock.json otherwise we're going to keep getting non-deterministic builds between development machines and deployment environments.


atharmon (approved):
👍

PR #261, opened 2022-07-14
feat(common-nest-services): Initial boilerplate

atharmon:
Generated boilerplate for:
- `libs/common/nest/services`

### DIFF
--- a/libs/common/nest/services/README.md
+++ b/libs/common/nest/services/README.md
@@ -1,7 +1,65 @@
-# common-nest-services
+# TAMU GISC NestJS Services
 
-This library was generated with [Nx](https://nx.dev).
+This library contains a module and a service that can be used in NestJS applications to simply and reliably send emails without having to implement `nodemailer` each time.
 
-## Running unit tests
+## Library Scope
 
-Run `nx test common-nest-services` to execute the unit tests via [Jest](https://jestjs.io).
+`@tamu-gisc/common/nest/services`
+
+### Modules
+
+- CommonNestServicesModule
+
+### Services
+
+- MailerService
+
+### Required environment variables
+
+- mailroomUrl: string
+
+## Setup
+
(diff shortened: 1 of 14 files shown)


Proantagonist:
What's the rationale between putting this in a standalone lib `common-nest-services` as opposed to exporting the service from the `@tamu-gisc/mailroom` scope?

libs/common/nest/services/src/lib/mailer/mailer.service.ts
```diff
@@ -0,0 +1,21 @@
...
+import { EnvironmentService } from '@tamu-gisc/common/nest/environment';
+import { IMailroomEmailOutbound } from '@tamu-gisc/mailroom/common';
+
+@Injectable()
+export class MailerService {
+  constructor(private readonly env: EnvironmentService) {}
+
+  public sendMail(mailConfig: IMailroomEmailOutbound) {
+    // TODO: Need the environment service to get the address of the mailroom-nest instance from environment file
+    const mailroomUrl = this.env.value('mailroomUrl');
```
Proantagonist, line 14: Is environment service expected to be loaded on the host application? I noticed there's no environment registered on the mailer module.

libs/common/nest/services/README.md
```diff
@@ -0,0 +1,7 @@
+# common-nest-services
```
Proantagonist, line 1: Is documentation going to be added in another PR?

Proantagonist (commented):
Only thing I would suggest is keep in mind the general scope of this library. 

What *all* is going to fall under the `common/nest/services` umbrella? It's a very broad scope.

 Would it not be better to have a more specific library `common/nest/utils/mailer`, for example?

libs/common/nest/services/src/lib/mailer/mailer.service.ts
atharmon: At the moment yes. I would like to integrate the EnvironmentModule in a coming PR so all you have to do is provide the environment variable within the app.module and not worry about importing EnvironmentModule yourself.

Proantagonist (approved):

PR #264, opened 2022-07-14
feat(mailroom-nest): mailroom-nest now should work on dev vm

atharmon:
Changed the code to use `relay.tamu.edu` or `smtp-relay.tamu.edu` based on an environment value. We should now be able to use `mailroom-nest` in the dev VM whereas before it didn't work because it was not a part of the TAMU network (or something like that).


Proantagonist (approved):
🚀 🚀 🚀

PR #267, opened 2022-07-22
feat(ues-operations-angular): Changed fiber paths layer to load from UES hosted service

Proantagonist:
This implements the SSO authorization for the IT fiber layers using the UES portal authentication.


apps/ues-operations-angular/src/environments/definitions.ts
```diff
@@ -3,6 +3,7 @@ import { SearchSource, SearchSourceQueryParamsProperties } from '@tamu-gisc/ui-k
 import { LayerSource } from '@tamu-gisc/common/types';
 
 import esri = __esri;
+import LayerSearchSource from 'esri/widgets/Search/LayerSearchSource';
```
atharmon, line 6: Orphaned import?

Proantagonist: Fixed

atharmon (approved):

PR #268, opened 2022-07-25
feat(oidc-provider-nest): Integration of MailService

atharmon:
This PR handles the integration of the MailService (created in PR #261) into the `oidc-provider-nest` application.

### DIFF
--- a/libs/oidc/common/src/lib/services/user/user.service.ts
+++ b/libs/oidc/common/src/lib/services/user/user.service.ts
@@ -4,8 +4,8 @@ import { hash, compare } from 'bcrypt';
 import * as deepmerge from 'deepmerge';
 
 import { EnvironmentService } from '@tamu-gisc/common/nest/environment';
+import { MailerService } from '@tamu-gisc/common/nest/services';
 
-import { Mailer } from '../../utils/email/mailer.util';
 import { SHA1HashUtils } from '../../utils/security/sha1hash.util';
 import { TwoFactorAuthUtils } from '../../utils/security/twofactorauth.util';
 import {
@@ -43,7 +43,8 @@ export class UserService {
     public readonly passwordResetRepo: UserPasswordResetRepo,
     public readonly passwordHistoryRepo: UserPasswordHistoryRepo,
     private readonly httpService: HttpService,
-    private readonly env: EnvironmentService
+    private readonly env: EnvironmentService,
+    private mailerService: MailerService
   ) {}
 
   public async insertDefaultAdmin() {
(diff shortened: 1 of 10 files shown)


libs/oidc/common/src/lib/services/user/user.service.ts
```diff
@@ -129,7 +130,13 @@ export class UserService {
...
     const newUser = await this.userRepo.save(user);
     if (newUser) {
-      Mailer.sendAccountConfirmationEmail(newUser.email, newUser.guid);
+      this.mailerService.sendMail({
+        to: newUser.email,
+        from: '',
+        subject: 'Your newly created GeoInnovation Service Center account',
+        text: 'An account for GeoInnovation Service Center has been created with this email address.',
+        html: `An account for GeoInnovation Service Center has been created with this email address. </br><a href="http://localhost:4001/user/register/${newUser.guid}">Verify email</a>`
+      });
```
Proantagonist, line 139: Should this be `await`ed?

libs/oidc/common/src/lib/services/user/user.service.ts
```diff
@@ -436,7 +443,15 @@ export class UserService {
...
-    Mailer.sendPasswordResetConfirmationEmail(user.email);
+
+    this.mailerService.sendMail({
+      from: '',
+      to: user.email,
+      subject: 'GeoInnovation Service Center password reset',
+      text: 'Your password to GeoInnovation Service Center has been reset.',
+      html: 'Your password to GeoInnovation Service Center has been reset.'
+    });
+
```
Proantagonist, line 454: Same thing here. Should it be awaited?

libs/oidc/provider/src/lib/controllers/interaction/interaction.controller.ts
```diff
@@ -136,7 +141,16 @@ export class InteractionController {
...
+
+        this.mailerService.sendMail({
+          to: user.email,
+          from: '"GISC Accounts Team" <giscaccounts@tamu.edu>',
+          subject: 'Hello from the IdP',
+          text: `Your two-step verification code is: ${token}`,
+          html:
+            `<p>Your two-step verification code is: <b>${token}</b></p>` +
+            `<p>Use this code to complete loggin in with GISC</p>`
+        });
```
Proantagonist, line 153: Should it be awaited? All of these emails are being sent out of order and CAN result in conditions where operations before or after the mail send fail and will either send an email regardless or not send one at all. Example:

1) View for 2fa confirmation gets rendered
2) Email with 2fa code doesn't get sent because of an error in 2fa code generation or email sending itself.

Proantagonist (changes requested):

Proantagonist (approved):

Issue #269, opened 2022-07-27
oidc-provider-nest: User sign up causes internal server error


atharmon:
Whenever you click the sign up link on the login menu the user is directed to /user/register and will cause the server to return **{"statusCode":500,"message":"Internal server error"}**. On the server itself the error message **Error: Unknown authentication strategy "jwt"** is presented. This is because the `UserController` is using a controller level guard `AdminGuard`. This needs to be moved into a route level guard where appropriate.

atharmon:
Addressed in PR #289

PR #289, opened 2022-08-10
fix(oidc-provider-nest): Changed AdminGuard scope in user.controller

atharmon:
Removed the controller level `AdminGuard` and instead put the guard on those routes that truly needed to be protected. 

Addresses issue #269.


Proantagonist (approved):
🚀 🚀 🚀

Issue #235, opened 2022-05-17
gisday-ngx-data-access: Remove LoginGuard


atharmon:
We can probably remove this and just use the `angular-auth-oidc-client`'s `AuthLoginPartialRoutesGuard` instead.

PR #291, opened 2022-08-11
fix(gisday-angular): Removed own LoginGuard from monorepo

atharmon:
I had created a `LoginGuard` inside of `gisday-ngx-data-access` prior to us discovering the much nicer `AutoLoginPartialRoutesGuard` from the `angular-auth-oidc-client` library. This PR removes the `LoginGuard` I created and replaces its usage with `AutoLoginPartialRoutesGuard`.

Addresses issue #235


Proantagonist (approved):
👍

PR #301, opened 2022-08-29
feat(kissingbug-angular): mailroom-nest integration

atharmon:
Updated `kissingbug-angular` to utilize the custom `emails` route on the Strapi backend; this works by POSTing the form from the Contact page to the `emails` route. This `email` controller will then forward the request to the appropriate `mailroom-nest` instance (dev, staging, prod).

Also cleaned up some code such as removed console.logs and whatnot.


Proantagonist (approved):
👍

PR #305, opened 2022-10-20
feat(gisday-bridge-nest): Initial commit of boilerplate

atharmon:
Decided to integrate my bridge conversion application into the monorepo so I don't need to maintain two different sets of GIS Day entities as this was proving to be quite the headache. 

Created:
- app/gisday-bridge-nest
- lib/gisday/bridge


Proantagonist (approved):
🚀

PR #347, opened 2023-09-28
ci: Add github actions workflows

Proantagonist:
Include gisday-angular and gisday-nest apps as test-bed.


Proantagonist:
Changes are so intertwined with the gisday-auth0, I'm just going to leave them in that branch

PR #444, opened 2025-09-30
Fix turf imports and lodash typings for Aggiemap build

lionwilliamson:
Fixed incorrect @turf/helpers import (moved Feature/Point to geojson types).

Aligned @types/lodash version with project’s TypeScript.

Verified build and run at http://localhost:4200.


Proantagonist:
Hi! Thanks for your submission. Can you provide some more details on the motivation for this PR? 

I'm noticing a several new dependency additions without any implementation details to suggest their need.

lionwilliamson:
Howdy!
When setting up AggieMap locally, the build failed due to missing type
dependencies ***@***.***/helpers and lodash). This fix adds the correct type
packages and updates the configuration so the project compiles
successfully. While the original codebase works for existing environments,
new contributors running a fresh install encounter errors that block them
from building and testing. This change ensures the repository is easier to
set up and maintain going forward.

Proantagonist:
Are you running in a dev container or bare-metal? The preferred method is dev container because it contains a sandboxed  environment configuration for deterministic dev builds. In any case, provided you have a compatible NodeJS version (v20) and system architecture (x86), `npm ci` should install all the correct dependencies; they're the exact same used in CI so builds do work with the current defined dependencies.

Example: https://github.com/TamuGeoInnovation/Tamu.GeoInnovation.js.monorepo/actions/runs/17952395119/job/51055094864

Dependencies are largely managed by the Nx workspace so we are fairly strict about dependency additions that are not strictly necessary because they have the potential to break/cause issues with peer dependency resolution with Nx workspace upgrade migrations.

Furthermore, while PNPM has benefits, particularly in monorepo's, this project currently uses NPM and the additional lock file might create issues and/or desync over time.

Please verify you're on a supported version of NodeJS and NPM and any relevant errors; I would like to resolve your dev env without requiring additional dependencies.

Proantagonist:
Closing due to inactivity. If this is a mistake, please reply to the above.

PR #522, opened 2025-11-11
feat(ts-events-angular): GIS Day event + parking map

sloneearly:
Adds a GIS day map tab with dates for GIS Day and important locations (MSC, parking garages, etc.) highlighted.

Replaces #516

### DIFF
--- a/libs/ts/events/ngx/src/lib/definitions/all.definitions.ts
+++ b/libs/ts/events/ngx/src/lib/definitions/all.definitions.ts
@@ -15,6 +15,7 @@ import { RingDayEvent } from './ring-day.definitions';
 import { ISpecialEventRoot } from '../interfaces/special-event.interface';
 import { MensBasketball_Ts } from './mens-basketball.definitions';
 import { WomensBasketball_Ts } from './womens-basketball.definitions';
+import { GisDayTs } from './gis-day-map.definitions';
 
 export const EventDefinitions: Array<ISpecialEventRoot> = [
   FourHRoundupTs,
@@ -32,5 +33,6 @@ export const EventDefinitions: Array<ISpecialEventRoot> = [
   TroubadourTs,
   FootballParkingEvent,
   RingDayEvent,
-  WomensBasketball_Ts
+  WomensBasketball_Ts,
+  GisDayTs
 ];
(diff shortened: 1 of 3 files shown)


Proantagonist:
Yo this is awesome! Your'e definitely getting the hang of the process :)

Nearly perfect:

1) Please make sure you have Prettier extension + configured as your default formatter. I believe we set this up during a call but the setting may not have stuck? There are some new-lines that were removed that will cause white-space changes the next time someone touches the file and/or runs a project-wide formatting with `npx nx run affected:lint`. This causes confusion in PR's because it introduces "changes" that aren't really changes.

2) Consider defaulting the zoom and center so the parking garages are within view. Without this, users will not know there's more to see past the default zoom, especially those far off garages like NSG.

Lastly, it's rarely needed to close another PR and re-open. Good intuition, picking up on what i would have said about that. Reminder that `git rebase -i <base_branch>` will allow you to drop commits you don't want to include and you can thereafter use `git push --force-with-lease`. Feel free to ping me next time you run into this, it can be scary since you can very easily accidentally prune good commits. Rebase is the swiss army knife of the git world; it's pretty awesome.

sloneearly:
> Yo this is awesome! Your'e definitely getting the hang of the process :)
> 
> Nearly perfect:
> 
> 1. Please make sure you have Prettier extension + configured as your default formatter. I believe we set this up during a call but the setting may not have stuck? There are some new-lines that were removed that will cause white-space changes the next time someone touches the file and/or runs a project-wide formatting with `npx nx run affected:lint`. This causes confusion in PR's because it introduces "changes" that aren't really changes.
> 2. Consider defaulting the zoom and center so the parking garages are within view. Without this, users will not know there's more to see past the default zoom, especially those far off garages like NSG.
> 
> Lastly, it's rarely needed to close another PR and re-open. Good intuition, picking up on what i would have said about that. Reminder that `git rebase -i <base_branch>` will allow you to drop commits you don't want to include and you can thereafter use `git push --force-with-lease`. Feel free to ping me next time you run into this, it can be scary since you can very easily accidentally prune good commits. Rebase is the swiss army knife of the git world; it's pretty awesome.

Thank you for the feedback, I have no idea what happened with the prettier extension, but I think I got it working now. Let me know if it still doesn't look right. I will be more diligent with that. And I changed the default zoom and center to show everything available on the map. 

Also, sorry for the weirdness with the PR's - to be honest I got so lost with the original one that I figured it would be easier to make a new one haha. But now I see how to rebase properly, so I will definitely do that in the future.

Issue #552, opened 2025-11-26
aggiemap-angular: superclass events structure to support variable sub-map types via routes and configuration files


danielgoldbergtamu:
### Issue

It has been requested that the Aggie Map team refocus our efforts to bring in the full set of parking maps available from the [Transportation Services dropdown-based map](https://transport.tamu.edu/parkingmap//tsmap.htm) in-line with or instead of our efforts to implement the rest of the sports-specific parking maps (which are listed in the same dropdown list on the same TS map page).

<img width="877" height="886" alt="Image" src="https://github.com/user-attachments/assets/fd30f12e-4be1-427d-a201-465628f92a64" />

### Current State

Presently, we have the ability to configure specific sub-maps using an approach where each sub-map gets its own configuration file and is displayed within the "Discover" section of the site and within the toast (pop-up) in the case that the event is upcoming or ongoing. 

<img width="2021" height="1225" alt="Image" src="https://github.com/user-attachments/assets/b77f11a8-ebb5-4053-901f-c4372645fa91" />

<img width="250" height="389" alt="Image" src="https://github.com/user-attachments/assets/9035e664-3661-46a5-aa94-283b59e873cf" />

<img width="540" height="871" alt="Image" src="https://github.com/user-attachments/assets/361e1bd0-74f6-4463-9ed3-7a6897b36e9f" />

### Proposed Plan

We can extend and repurpose the existing event structure where sub-maps are defined via configuration files to support addition application routes with custom maps defined within each route. 

The **/events/** portion (route) of the url https://aggiemap.tamu.edu/events/womens-basketball/map/d can be extended to support **/parking/** or **/anything/** and pick up maps that are within that route.

This would look like **/parking/avp/** in the following ultimate URL, with AVP being a map configuration of its own like events are now to support the Any Valid Permit parking layers presently displayed in the [TS AVP Map](https://transport.tamu.edu/parkingmap/tsmap.htm?map=avp)

- https://aggiemap.tamu.edu/parking/avp/map/d 

<img width="1293" height="1107" alt="Image" src="https://github.com/user-attachments/assets/983b1059-7a10-438d-8aba-a5c5ff2b9192" />

PR #573, opened 2026-01-13
feat(aggiemap-angular): Add TS Move Out Map

daltonpetersn:
part of #552

Built out configuration for TS Move Out Map


.vscode/settings.json
```diff
@@ -37,5 +37,6 @@
   "angular.enable-strict-mode-prompt": false,
   "cSpell.words": [
     "Teague"
-  ]
+  ],
+  "angular.log": "verbose"
```
Proantagonist: What is the purpose of this?

PR #586, opened 2026-01-15
feat(aggiemap-angular): Add TS Contractor Parking Map

daltonpetersn:
Added definitions for Transportation Services Contractor Parking Map

Part of #552


daltonpetersn:
Closed to determine if definitions and commits were made on the correct branches. Reopened once verified.

Issue #593, opened 2026-01-27
aggiemap-angular: confirm that all semester events are created


danielgoldbergtamu:
### Issue

We have been asked to confirm that all of the important events for the semester have had configurations created and updated for the new semester, and to make sure that all of the events work properly. 

### Some things to check:

1. The events are all created
2. The events have had their configuration dates updated for the new semester
3. The underlying data layers from ArcGIS have had their dates updated, meaning that the data for the map show properly and that the data layers are not still using the old dates from the prior year.

### List of events to verify

1. Baseball opening day, multiple event conflict - February 13
2. Aggieland Saturday - February 28
3. Baseball/Basketball conflict - Friday, March 3
4. Big Event - March 21
5. Physics and Engineering Festival - March 28 (tentative as of 1/8)
6. Maroon and White game - April?? Date TBD
7. Ring Day - April 9-10
8. Family Weekend - April 10-12
9. Muster - April 21
10. MS 150 - April 26
11. Savannah Bananas - May 2
12. Graduation - May 7-9

### What to do if something is wrong

**If any of the events have incorrect ArcGIS service layers:**

1. Email Megan (megan_mcmullen@tamu.edu) and Eric (esirwin@tamu.edu) to get clarity and/or correction to the data layers. This will usually be that the dates from last year have not been updated in the data layers underneath the ArcGIS services.

**If any of the events do not already have configurations created:**

1. Email Megan (megan_mcmullen@tamu.edu) and Eric (esirwin@tamu.edu) to get clarity on the dates and to determine which data layers should be included along with an example image of the what the map is supposed to look like.

sloneearly:
This is ready to be closed.

PR #595, opened 2026-02-02
feat(aggiemap-angular): Add Savannah Bananas parking map

sloneearly:
Built config file for Savannah Bananas baseball parking, uses map layer from Maroon and White game with banana-themed colors.

Part of #593 

Screenshot 2026-02-02 111115


sloneearly:
Not sure what the parking lot info (i.e. parking permit prices) needs to be yet as we do not have the necessary information. 

Also, I will change the colors if needed. Just thought it would be fun.

Proantagonist:
> Not sure what the parking lot info (i.e. parking permit prices) needs to be yet as we do not have the necessary information.
> 
> Also, I will change the colors if needed. Just thought it would be fun.

Perfect. I like the creativity, we'll keep the colors for now ;)

PR #636, opened 2026-02-20
fix(aggiemap-angular): Update Ring Day Dates + Fix Issue with Ring Day Map Visibility

sloneearly:
This PR fixes two labels that were missed when initially updating the maps to 2026. 

The service data dates don’t seem to match 2026 yet, which caused date filters to return zero results and show an empty map. This PR replaces date-based layer expressions with 1=1 for Areas, Routes, and POIs to ensure everything shows.

### DIFF
--- a/libs/ts/events/ngx/src/lib/definitions/ring-day.definitions.ts
+++ b/libs/ts/events/ngx/src/lib/definitions/ring-day.definitions.ts
@@ -140,11 +140,11 @@ export const RingDaySpecialEventOptions: SpecialEventOptions = [
     choices: [
       {
         value: EventDay.DAY1,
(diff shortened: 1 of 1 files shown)


Proantagonist:
So if it's always 1=1 for all event dates, then there's no point in having multiple days? @sloneearly

sloneearly:
> So if it's always 1=1 for all event dates, then there's no point in having multiple days? @sloneearly

Good point. My thinking was that since they were the same map the dates wouldn't matter, but now it has been requested that April 8th be added, so I see how it is supposed to work now. I just pushed a new commit that compares the dates and filters like before, but with three days now.

PR #639, opened 2026-02-25
feat(aggiemap-angular): New Move In Builder

sloneearly:
This PR implements a new builder system for the Move In map. It is the same style as the old Move In builder and will function as a template for future builders on large/feature heavy maps.

Functional and ready for development, but currently waiting to know if layers are filtering correctly or if any changes need to be made.

image

image

image

image

image

image


danielgoldbergtamu (approved):
Looks good.

PR #644, opened 2026-03-04
fix(aggiemap-angular): Manually set map center and zoom for Big Event

sloneearly:
This PR resolves an issue where the Big Event map was centered in an incorrect location in the ocean:
image

This PR now centers it around Reed Arena with a default zoom level of 17.

image


danielgoldbergtamu (approved):
Minor changes. Seems fine for push to dev for testing.

meganmcmullen-rgb:
Can I get the Traffic Flow layer moved to the top? That way the closures and arrows are visible if they are overlapping lots. Thanks!

sloneearly:
> Can I get the Traffic Flow layer moved to the top? That way the closures and arrows are visible if they are overlapping lots. Thanks!

Done!

PR #645, opened 2026-03-04
fix(aggiemap-angular): Remove unnecessary NSC and Break/Summer Parking Map duplicate files

sloneearly:
Removes the following files that are duplicates/outdated versions of the NSC and Break/Summer configuration files:

break_summer.definitions.ts
break_summer.ts
nsc-parking-definitions.ts


danielgoldbergtamu (approved):
Looks good.

PR #647, opened 2026-03-06
fix(aggiemap-angular): Reordered traffic layers to top

sloneearly:
Puts traffic layers on top of other layers, as requested by Megan (@meganmcmullen-rgb in https://github.com/TamuGeoInnovation/Tamu.GeoInnovation.js.monorepo/issues/644#issuecomment-4005876905_

image


danielgoldbergtamu (approved):
Looks good.

PR #652, opened 2026-03-16
fix(aggiemap-angular): Fix issue with safety-first on sport map legends

sloneearly:
This PR fixes issues with the legend functionality for the new sporting maps. Somewhere in the chaos of dealing with the merge conflicts, the previously fixed issue with the safety-first layer not properly appearing on the legend reappeared. 

### **Changes:**

- `legend-element.component.ts`
Updated useSportsSafetyFirstFallback() to recognize Safety First layers in two ways: directly by the parent MapImageLayer ID, and indirectly when the layer passed in is a child sublayer of one. Also removed the requirement for the element to have zero info. Safety First layers now always use the fallback regardless.

- `legend-element.component.html`
Wired up the fallback rendering that was already written in the component but never used in the template. Moved it before the element type switch so it takes priority over all element types. Previously it only fired in the "unknown type" default case, meaning symbol-table elements slipped through and showed "Safety First" as the label.

- `legend-collection.component.ts`
Added the Safety First layer IDs to this component as well. Used them in two places: showGroupTitle suppresses the bold "Safety First" group header that was appearing above the legend entry, and visibleChildGroups / _shouldHideSportsSafetyFirstChildGroups() prevents duplicate rendering when the parent layer has its own legend elements alongside child sublayer groups.

- `legend-collection.component.html`
Swapped hasChildren for showGroupTitle on the group title header, and childGroups for visibleChildGroups on the child collection loop, applying the two new conditions from the component.

Before:
Screenshot 2026-03-16 162449

After:
image

### DIFF
--- a/libs/maps/feature/legend/src/lib/components/legend-collection/legend-collection.component.html
+++ b/libs/maps/feature/legend/src/lib/components/legend-collection/legend-collection.component.html
@@ -1,7 +1,7 @@
 <div class="legend-collection">
   <div
     class="collection-group-title"
-    *ngIf="hasChildren"
+    *ngIf="showGroupTitle"
     tabindex="0"
     role="button"
     (click)="toggleExpanded()"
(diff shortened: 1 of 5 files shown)


danielgoldbergtamu (approved):
Seems fine.

PR #653, opened 2026-03-18
fix(aggiemap-angular) Put main campus parking back in all.definitions file

sloneearly:
Main campus parking got removed from the definitions registry file and thus stopped showing up on Aggiemap. This PR adds it back.


danielgoldbergtamu (approved):
Looks good.

PR #658, opened 2026-03-18
fix(aggiemap-angular): Reorder Physics Festival layers and fix issues with visibility toggle

### LINKED ISSUE #655: (aggiemap-angular): Physics Festival layer order and visibility updates
The dropdown menus are no longer affecting the visibility of the Physics Festival layers:

Image

From Megan:

> Can we have it set to General Parking as the default on when the map opens and then leave the bus layers off? And would it be possible to bring those 3 layers to the top in the layer list so they don't get lost? They're mixed in with the other default layers and a bit hard to find.

Image

Proposed Solution:

Fix: Add the following to PhysicsAndEngineeringFestivalConfiguration:

```
legendAllowVisibilityToggle: true
legendCombineChildrenUnderPrimary: true
legendExcludedLayerIds: [...] (same list already in EventLegendComponent._physicsExcludedLayerIds)
```
In PhysEngFestivalColdLayerSources, change visible on the two bus group layer sources:

```
EASTBOUND_UNIVERSITY_GROUP group: visible: false
WESTBOUND_UNIVERSITY_GROUP group: visible: false
GENERAL_PARKING_GROUP group: remains visible: true
```

Use defaultLayerOverrides in PhysicsAndEngineeringFestivalConfiguration to set listMode: 'hide' on the default aggiemap layers that are not relevant for this event (e.g., parking, dining, POI, construction zones, etc.). This reduces noise in the layer list, making the 3 Physics Festival group layers easier to find.

Investigate/verify the layer addition order in the map service/event settings service to ensure Physics Festival group layers are added to the ESRI map before the default layers, which would position them at the top of the layer list. If the order is already correct but the ESRI layer list is sorting them differently, a listMode override on the default layers (step 1) may be sufficient on its own.

sloneearly:
This PR sets the General Parking as the default layer for the Physics festival maps and fixes an issue with the toggle-able visibility from the dropdowns not working.

Before:

image

After:
image

Closes #655

### DIFF
--- a/libs/ts/events/ngx/src/lib/definitions/phys-engineering-festival.definitions.ts
+++ b/libs/ts/events/ngx/src/lib/definitions/phys-engineering-festival.definitions.ts
@@ -213,7 +213,7 @@ export const PhysEngFestivalColdLayerSources: LayerSource[] = [
     type: 'group',
     id: PhysEngFestDefinitions.EASTBOUND_UNIVERSITY_GROUP.id,
     title: PhysEngFestDefinitions.EASTBOUND_UNIVERSITY_GROUP.name,
-    visible: true,
+    visible: false,
     listMode: 'show',
     sources: [
       {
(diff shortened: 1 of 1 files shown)


danielgoldbergtamu (approved):
Looks good.

PR #660, opened 2026-03-18
fix(aggiemap-angular): Remove unnecessary reserved parking layers and updated volleyball on sports maps

### LINKED ISSUE #654: (aggiemap-angular) Remove reserved parking lots from sports parking maps + volleyball map edit
From Megan:

> Sports maps - I'm going to hold off on republishing all these for now, just because I'd rather not redo it again once the pull back is finished.
Softball looks good
Soccer - I'm not sure if I need to republish, but we don't have any reserved parking so it will need to be removed from the legend.
Swimming - same as soccer, we don't need reserved parking (I may need to republish I'm not sure).
Volleyball - Reserved Parking should be Accessible Parking Only
Tennis - looks good
Outdoor track - looks good
Cross Country - needs reserved parking removed from the legend

Example:

Image

Proposed Solution:

Remove the reserved parking layers and associated code from the configuration files for each of the affected sport parking maps. Change name of reserved parking layer on the Volleyball map to Accessible Parking Only.

sloneearly:
This PR removes the Reserved Parking layers from Soccer, Swimming, and Cross Country, and updated the Reserved Parking on Volleyball to "Accessible Parking Only."

Close #654

### DIFF
--- a/libs/ts/events/ngx/src/lib/definitions/cross-country-parking.definitions.ts
+++ b/libs/ts/events/ngx/src/lib/definitions/cross-country-parking.definitions.ts
@@ -46,12 +46,6 @@ const crossCountryAreaRenderer: FeatureRenderer = {
 const crossCountryParkingRenderer: FeatureRenderer = {
   type: 'unique-value',
   field: 'GIS.TS.SPEV_Lot_Use.TrackXC',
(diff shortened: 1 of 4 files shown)


danielgoldbergtamu (approved):
Looks good.

PR #664, opened 2026-03-20
fix(aggiemap-angular) Combine Sustainable Transporation bike rack layers

### LINKED ISSUE #661: (aggiemap-angular): Combine all bike racks into one layer
It has been requested that all bike racks be combined into a single layer, rather than all being unique layers:

From Megan:

> Sustainable - Since the racks all have the same symbol on the map, is there a way to label them as "Bike Racks" on the legend instead of having the specific type of rack? That would shorten the legend and I think it would fit on one screen.

Image

sloneearly:
This PR combines the numerous bike rack types under Sustainable Transporation into one bike rack layer:

### Before:
image

### After:
image

Closes #661


danielgoldbergtamu (approved):
Looks good.

PR #667, opened 2026-03-20
feat(aggiemap-angular): Skip unnecessary builder on maps with no selection options

### LINKED ISSUE #662: (aggiemap-angular): Remove the builder from parking and events that do not have builder selections
### Issue

1. The builder interface is shown by default in all maps for parking and events that have been created using the new config specification. 
2. This does not make sense when the special map has no options to select. 
3. The builder should be removed/hidden for event/parking maps where there is nothing to select or configure.
4. Links to event/parking maps should take the user straight to the map and not force them to click through the builder first.

### Example: Any Valid Parking

The [Any Valid Parking](https://aggiemap.tamu.edu/parking/avp-parking/builder/intro) map does not have any selections that a user could or should make, so it does not need a builder. 

Image

sloneearly:
This PR updates builder routing so maps with no configurable builder options open directly to the map, while maps with actual builder options remain unchanged.

## Changes:

- Added an `EventEntryGuard` to handle bare event routes like `/:eventId`
- If an event has no builder options, the user is sent straight to `/map`
- If an event has configurable options, the user is sent to `/builder/intro`
- Added a `BuilderAccessGuard` so intro-only events cannot be forced back into the builder route
- Kept all existing multi-step builder flows intact for maps that actually require configuration
- Future map configs should be handled automatically by the logic in event-entry.guard.ts, builder-access.guard.ts, and event-settings.service.ts

Closes #662

### DIFF
--- a/libs/ts/events/ngx/src/lib/guards/builder-access/builder-access.guard.spec.ts
+++ b/libs/ts/events/ngx/src/lib/guards/builder-access/builder-access.guard.spec.ts
@@ -1,0 +1,59 @@
+import { TestBed } from '@angular/core/testing';
+import { ActivatedRouteSnapshot, Router, UrlSegment, UrlTree } from '@angular/router';
+import { RouterTestingModule } from '@angular/router/testing';
+
+import { BuilderAccessGuard } from './builder-access.guard';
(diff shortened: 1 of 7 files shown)


danielgoldbergtamu (approved):
Nice.

PR #676, opened 2026-03-20
feat(aggiemap-angular): Update toasts for Big Event, Baseball, and Physics Festival

### LINKED ISSUE #674: aggiemap-angular: event toasts not working
### Issue

It seems that the event toast (popup's about the events) stopped being shown on the main Aggie Map site.

### Example

Big Event is this weekend, but the toast does not load on the Aggie Map homepage

Image

Image

### Resolution

1. Update and/or verify the config files for each of the events to make sure that the config data that specifies the popup is there and has the correct dates and other data so that the toasts will show for the events.

sloneearly:
This PR adds toast notifications for Big Event, Baseball, and Physics Festival.

Screenshot 2026-03-20 134112

Closes #674


danielgoldbergtamu (approved):
Great.

PR #680, opened 2026-03-23
fix(aggiemap-angular): Hide bananas 

### LINKED ISSUE #678: aggiemap-angular: Remove "Savannah Bananas Parking" map from aggiemap
### Issue

The Savannah Bananas parking map service does not exist yet, it has not been finalized or published so we need to remove this from AggieMap until I publish the correct version.

### Resolution

1. Remove map from aggiemap and under athletic events until the service has been published

### Notes

When this map is created it will need to be under the campus events heading.

sloneearly:
This PR hides the Savannah Bananas parking map from the discover page. The map configuration file is built out, but we do not have the actual parking map service yet. When we do have it, it can be inserted into the config file.

Closes #678


danielgoldbergtamu (approved):
Good.

PR #682, opened 2026-03-23
fix(aggiemap-angular): Direct Discover map builder to accommodations page instead of intro page

### LINKED ISSUE #675: aggiemap-angular: Remove "Get Started" page button from event builders
### Issue

Users report that there it takes too many clicks to get through the builder pages and that the "Get Started" button could safely be removed.

Image

### Resolution

**Using Big Event as an example:** 

- Reorganize the builder processing to not require clicking through the ["Get Started" section](https://aggiemap.tamu.edu/events/big-event/builder/intro) 

- Instead go straight to the [first option selection page](https://aggiemap.tamu.edu/events/big-event/builder/accommodations/map-type).

Image

This would then need to be replicated across all builders to produce the same experience. 

### Note

It may be or should be possible to do this at a parent level of the component workflow, rather than at each of the configuration levels.

sloneearly:
This PR directs the Discover map builder to the accommodations page rather than the intro "Get Started" page. 

Functionality for the Get Started page still exists in the codebase, but it will not be used anymore unless needed.

Closes #675

### DIFF
--- a/libs/ts/events/ngx/src/lib/guards/event-entry/event-entry.guard.ts
+++ b/libs/ts/events/ngx/src/lib/guards/event-entry/event-entry.guard.ts
@@ -26,7 +26,7 @@ export class EventEntryGuard implements CanActivate {
       return true;
     }
 
-    const targetCommands = this.eventSettingsService.hasOptions ? ['builder', 'intro'] : ['map'];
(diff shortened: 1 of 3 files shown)


danielgoldbergtamu (approved):
Good

PR #691, opened 2026-03-25
feat(aggiemap-angular): Add bus stop popups to Main Campus Parking map + labels to Start/Stop Points + extend Popup Component behavior

### LINKED ISSUE #684: (aggiemap-angular): Campus Main Parking map needs symbology and popups for bus stops
### Issue 1:
The Campus main parking map should have a popup when the bus stop is selected that says the stop name and route numbers (from StopName & Route in the attributes) along with a link to the routes: https://aggiespirit.ts.tamu.edu/RouteMap that is titled "view on the bus route map"

Image

### Issue 2:
The main bus stops on the map should be labeled on the map:

Image

sloneearly:
This PR adds customized popups to each bus stop of the Main Campus Parking map and adds number labels to each Start/Stop point, just as it was on the Campus Main Parking Map on Transporation Services. This PR also extends shared popup behavior with optional content instead of introducing a special-purpose popup component.

### Changes

- Updated the Main Parking Campus Stops layer to use MarkdownPopupComponent instead of a dedicated `CampusStopsPopupComponent` (previously used inefficient and overengineered `CampusStopsPopupComponent` instead of general solution)
- Created new reusable popup data support 
- Extended popup data typing to support literal config values via `{ value: string }`.
- Updated `PopupService` to resolve literal popup values without treating them as attribute lookups, while still supporting template rendering when needed.
- Extended both shared popup components to support optional 'additionalContent'
- Added `PipesModule` to support markdown parsing in the shared popup module.

  #### Note on Start/Stop points:
  
  - The original labeling approaches were not working for this layer because the ArcGIS source is Multipoint, which does not behave well with client-side `labelingInfo`, and the earlier definition-based label experiments were not part of the event layer loading flow. As a result, the labels never rendered consistently in the app.
  - To make this work , I moved the solution into the event runtime instead of the static layer definitions. In `event.service.ts`, the app now detects the `ts-main-parking` event after its layers load, queries the Route Stop/Start Points features from the TS Main service, expands each multipoint feature into individual point graphics, and adds a client-side `GraphicsLayer` directly above the route stop layer. Each graphic renders the route number as white, bold Open Sans text centered on the stop icon, with a small white halo to match the original map more closely.

### Before:
image

### After:
image

Closes #684


danielgoldbergtamu (changes requested):
I don't like this.

Instead of creating a one-off new pop-up type, it would have been better to extend the existing pop-up to allow for an optional extra text section that pulls its contents from the configuration for the event/parking map.

We prefer a general solution that a) is optional and does not affect or be visible to other components, b) does not introduce new components that are not re-usable.

libs/ts/events/ngx/src/lib/services/event/event.service.ts
danielgoldbergtamu, line 1: I'm not sure that putting layer specific information/symbology into this file is the correct location. @Proantagonist am I correct? Where should this go to not clutter up this mid-level file.

libs/ts/events/ngx/src/lib/modules/popups/popup-content.utils.ts
danielgoldbergtamu, line 1: I think utils like this should probably go into a more generic /utils location rather than just in line here. @Proantagonist thoughts?

libs/ts/events/ngx/src/lib/modules/popups/markdown-w-directions-popup/markdown-w-directions-popup.component.scss
danielgoldbergtamu, line 1: Please check that this works on mobile as well.

danielgoldbergtamu (changes requested):
Just want to check with @Proantagonist that this new code is in the correct places. It feels like it is not. 1. The styles are in the main event service, when no other styles for any other layers are. 2. There is likely another section that common utils are supposed to be placed in.

Proantagonist:
I'll follow up with more feedback later, but I have one big question regarding the rendering issues with multipoint geometry and labelingInfo, and I only ask because I've had to go this road several times before, but have we tried loading the layer in a new version of the JS API to see if it's suffering from rendering issues? If the problem is resolved, then that likely means our current version @ 4.23 has compatibility issues with that particular implementation and if its fixed in a newer version, then that'd be the cleanest solution. However, important to test because migrating to a much newer JS API verison requires that we migrate the whole mapping core to using ES Modules instead of AMD modules which require that we have the module provider, and that's a huge lift. @danielgoldbergtamu @sloneearly

libs/ts/events/ngx/src/lib/modules/popups/markdown-w-directions-popup/markdown-w-directions-popup.component.scss
sloneearly: Working on mobile.

PR #693, opened 2026-03-25
feat(aggiemap-angular): Group concurrent event toasts into single stacked notification

### LINKED ISSUE #687: aggiemap-angular: Multiple Toasts
### Issue

When two (or more) event toasts are active, the toasts stack on top of each other. 

In desktop, there is a slight cascade event so the user can see there are multiple, but in mobile only the foremost toast can be seen. 

Is there another way to visually represent multiple toasts?

**Affects:** Main map and Discover pages

Image

### Proposed Resolution

**Option 1:** If more than one event is active, have just one popup that says something like "Several upcoming events may affect campus parking or have special parking instructions. Click here to see upcoming events"? This would then take them to the list of events on discover?

**Option 2:** Have one popup with the ability for users to click through to the next one to see more events? Maybe have to start with a little baby lander popup text: "Several upcoming events may affect campus parking or have special parking instructions. Click here to see upcoming events"

sloneearly:
This PR fixes the issue where multiple active event notifications would render as overlapping toasts, making them unreadable on mobile.

### Changes

- Adds an optional `[grouped]` input to `tamu-gisc-notification-container`. When enabled, all active notifications are combined into a single stacked toast with separators between entries and individual "Don't show again" links. The grouped toast shares the same 10-second auto-dismiss timer, hover-to-pause behavior, and slide animation as individual toasts.
- The grouped flag defaults to false, so no other apps using the notification container are affected. AggieMap opts in via `[grouped]="true"` in `app.component.html`, covering both the main map and Discover pages.

### Before:
image

### After:
image

Closes #687

### DIFF
--- a/apps/aggiemap-angular/src/app/app.component.html
+++ b/apps/aggiemap-angular/src/app/app.component.html
@@ -1,3 +1,3 @@
 <router-outlet></router-outlet>
 
-<tamu-gisc-notification-container></tamu-gisc-notification-container>
+<tamu-gisc-notification-container [grouped]="true"></tamu-gisc-notification-container>
--- a/libs/common/ngx/ui/notification/src/lib/components/notification-container/notification-container.component.html
+++ b/libs/common/ngx/ui/notification/src/lib/components/notification-container/notification-container.component.html
@@ -1,8 +1,37 @@
-<tamu-gisc-notification-item 
-  *ngFor="let item of notifications | async" 
-  [notification]="item" 
-  (closeNotification)="close($event)" 
-  (acknowledgeNotification)="acknowledge($event)"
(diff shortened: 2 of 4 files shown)


danielgoldbergtamu (commented):
@Proantagonist Any idea what the build error "Error: bundle initial exceeded maximum budget." means?

danielgoldbergtamu (commented):
@Proantagonist Seem ok?

danielgoldbergtamu:
@sloneearly Can you look back into this once you get all of the other event stuff handled? We need to get this multi-toast issue resolved.

danielgoldbergtamu:
@sloneearly Can you look into this one again so we can get this wrapped up?

PR #700, opened 2026-03-27
fix(aggiemap-angular): Update athletic event names

### LINKED ISSUE #690: aggiemap-angular: Athletic Events - Wording
### Issue

There is better wording that we can use for some of the text on the Athletic Events section of the Discover Page.

Image

### Resolution

**Subhead:** In the subhead italic text, change "game day" to "gameday"

**Football:** Change the name of the "Gameday Transportation" map to just "Football"

**Parking** Most maps include the word "Parking" in the title, but they almost all contain info beyond parking.
1. Remove the word "Parking"
2. Replace "Parking" with "Map" (which may be a little redundant)

sloneearly:
This PR updates names for athletic events as well as some other text relating to athletic text on the discover page.

### Before:

image

### After:

image

Closes #690

### DIFF
--- a/libs/aggiemap/ngx/discover/src/lib/components/discover.component.html
+++ b/libs/aggiemap/ngx/discover/src/lib/components/discover.component.html
@@ -141,7 +141,7 @@
           aria-labelledby="discover-tab-athletics"
         >
           <h2>Athletic Events</h2>
-          <p class="section-description">Browse athletic event maps for game day parking and transportation information.</p>
+          <p class="section-description">Browse athletic event maps for gameday parking and transportation information.</p>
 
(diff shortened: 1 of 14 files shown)


danielgoldbergtamu (changes requested):
You added "Map" to all of them, but left "Football" without it ("Football Map"). 

If we are going to have "Map" on there it should be on all so that they are consistent. 

Add "Map" to "Football Map".

sloneearly:
> You added "Map" to all of them, but left "Football" without it ("Football Map").
> 
> If we are going to have "Map" on there it should be on all so that they are consistent.
> 
> Add "Map" to "Football Map".

Just fixed it @danielgoldbergtamu. I went ahead and updated Softball Regionals too. 

image

danielgoldbergtamu (approved):
Thanks.

PR #701, opened 2026-03-27
fix(aggiemap-angular): Make back button work on Aggiemap builder page

### LINKED ISSUE #689: aggiemap-angular: Event Builder, Back Arrow Does Not Work
### Issue

Clicking the back button in the browser does not return the user to the Discover page.

Image

sloneearly:
This PR adds `replaceUrl: true` to the automatic redirect in the builder so the intermediate URL (`builder/accommodations`) doesn't create a history entry that traps back navigation in a loop.

lower

Closes #689


danielgoldbergtamu (approved):
Nice.

PR #707, opened 2026-03-30
fix(aggiemap-angular): Fix SEC Grounds route and checkpoint icon proportions + Refactor Bike Rack legend

### LINKED ISSUE #705: aggiemap-angular: Icons on SEC Grounds Conference Map are stretched
The icons for certain layers on the SEC Grounds Conference map appear to be stretched:

Image

sloneearly:
This PR fixes stretched SEC Grounds Conference icons in both the map and legend.

### Changes:

- Added a 'createConferencePoiLayerNative' factory function builds the unique-value renderer for a given day number and is applied to the Day 2 and Day 3 POI layer sources. Day 1 POIs retain their default symbology as the data for that layer uses a different structure.

### Before:

image

### After: 

image

### Notes:

- The Day 2/3 checkpoint artwork is a tall pin, but the live SEC service publishes it as a square picture-marker symbol, which causes distortion on the map. (This also appears to happens on the Baseball map)

image

Closes #705


danielgoldbergtamu (changes requested):
I again don't like that these changes for working on a specific layer are in a major base component/style. Is it not possible to keep all of the changes needed for a fix for a specific layer closer to the definition file for the specific map rather than pushing them up into the core components where we do not have layer-specific stuff for anything else?

sloneearly:
@danielgoldbergtamu I will update the PR description with this info, but basically, I added a input called 'legendSrcOverrides' that replaces the SEC Grounds Conference specific code with an opt-in setting for custom icon rendering on the legend. So now the fix that was made for SEC Grounds Conference can be used for any map.

libs/maps/feature/legend/src/lib/components/legend-element/legend-element.component.spec.ts
sloneearly, line 1: Meant to say "Updated spec file" not "Updated scss file"

libs/maps/feature/legend/src/lib/components/legend-element/legend-element.component.html
```diff
@@ -42,6 +42,7 @@
                 [respectDefinitionExpression]="respectDefinitionExpression"
                 [hideGroupHeader]="hideGroupHeader"
                 [useBikeRackLegendTransform]="useBikeRackLegendTransform"
```
danielgoldbergtamu: Do we still need to have specific flags for using bike rack transforms in this base component?

libs/maps/feature/legend/src/lib/components/legend-element/legend-element.component.ts
```diff
@@ -128,7 +137,8 @@ export class LegendElementComponent implements OnInit {
     }
 
     const operableInfos = (this.element.infos ?? []) as Array<LegendInfo>;
-    const displayInfos = this.shouldApplyBikeRackLegendTransform ? this._getBikeRackLegendInfos(operableInfos) : operableInfos;
+    const bikeRackAdjustedInfos = this.shouldApplyBikeRackLegendTransform ? this._getBikeRackLegendInfos(operableInfos) : operableInfos;
```
danielgoldbergtamu: Same here. We would like to keep the base files generic, and not include variables, functions, or other things specific to specific layers in the generalized versions like this file.

libs/maps/feature/legend/src/lib/components/legend/legend.component.html
```diff
@@ -14,5 +14,6 @@
     [allowVisibilityToggle]="allowVisibilityToggle"
     [combineChildrenUnderPrimary]="combineChildrenUnderPrimary"
     [useBikeRackLegendTransform]="useBikeRackLegendTransform"
```
danielgoldbergtamu: Not needed any more now that you have an optional way of doing it?

sloneearly:
@danielgoldbergtamu Scratch what I said earlier - did not actually need any of the legend override functionality. All changes are now in the SEC Grounds Conference config file.

libs/maps/feature/legend/src/lib/components/legend/legend.component.ts
```diff
@@ -65,6 +68,7 @@ export class LegendComponent implements OnInit, OnDestroy {
         routeData['combineChildrenUnderPrimary'] ?? this.combineChildrenUnderPrimary;
       this.useBikeRackLegendTransform =
```
danielgoldbergtamu: Same as well. Layer specific things should not be in the general legend components. I think you have a way to do things like this in an optional approach with overrides?

danielgoldbergtamu (changes requested):

libs/maps/feature/legend/src/lib/components/legend/legend.component.ts
sloneearly: EDIT: I missed some more bike rack leftovers, currently going through them now

I believe this is a leftover from when the Bike Rack layer on sustainable transportation was split into many different types of bike racks. The bike rack layer still works at intended without `useBikeRackLegendTransform.` I went ahead and removed them in the [most recent commit](https://github.com/TamuGeoInnovation/Tamu.GeoInnovation.js.monorepo/pull/707/changes/1da260c6016511553996e064a4022a740cd7434a). 

image


I figured since its a small change, it was fine to remove them in this PR, but if you feel that would cause confusion, I can make a separate issue and PR.

Here's where and when `useBikeRackLegendTransform` was made originally for context: https://github.com/TamuGeoInnovation/Tamu.GeoInnovation.js.monorepo/pull/664/changes#diff-963d4eca4fcd471e62175a66248573299c1b990d5c12000eb0de90be1aa78170

sloneearly: Bike racks have been refactored and there is no longer anything in the higher-level files.

https://github.com/TamuGeoInnovation/Tamu.GeoInnovation.js.monorepo/pull/707/changes/3cff39d792534c6390b2e7d66e5f4355fd781cbf

libs/aggiemap/ngx/common/src/lib/layer-sources/layer-sources.ts
```diff
@@ -23,6 +23,9 @@ export function LayerSources(
 ): Array<LayerSource> {
   const bikeMapUrl = connections.tsMainUrl.replace('/TS_Main/MapServer', '/BikeMap/MapServer');
   const evChargeStationsUrl = connections.tsMainUrl.replace('/TS_Main/MapServer', '/EVChargeStations/MapServer');
+  const rackLayersUrl = 'https://arc.ts.tamu.edu/arcgis/rest/services/Hosted/Rack_Locations_view/FeatureServer';
```
danielgoldbergtamu: See if it is possible to move all of these types of arrays and sources into the main "all" one below so that we do not have specific logic for specific layers: 

const all: Array<LayerSource> = [

danielgoldbergtamu (changes requested):

danielgoldbergtamu (approved):
Yay. Good work.

PR #712, opened 2026-04-06
feat(aggiemap-angular): Add event details links to event builders and update Maroon & White Game map

### LINKED ISSUE #710: On Discover/event maps, replace dates with link to event website
The new link should direct users to the source website. For example, for athletics events, they should link to that sport's athletics page.

Image

### LINKED ISSUE #711: aggiemap-angular: Maroon & White Game map updates
### Issue

1. Maroon & White map needs to be updated, lots were reconfigured as seen below with popups that show descriptions (these are working on the icons already just need to make sure they are added for the lots). 

Layers:
Accessible/PrePaid Parking (0)
Event Parking Lots (1)

Image

2. One or both icons are stretched on Aggie Map

sloneearly:
This PR adds support for event-specific schedule/detail links in the event builder. It also updates the Maroon & White Game map with newly-formatted icons and a new parking map layer.

## Changes
- adds optional `scheduleUrl` to the shared event configuration interface
- update the events builder to show a clickable event details/schedule link when `scheduleUrl` is provided and builder is enabled
- adds matching link styling in the events and football builders
- populate `scheduleUrl` across event definitions forfootball, baseball, basketball, track, soccer, softball, volleyball, swimming, Family Weekend, Ring Day, Move-In, Big Event, graduation, and Savannah Bananas
    -  (May not be necessary for other athletic events that currently do not have builders, but I left these in case we want to reuse them and/or make a builder for these events in the future. Can be removed if desired.)
- updates the Maroon & White Game configuration date to April 18, 2026
- adds custom renderers and marker artwork for Maroon & White Game accessible and prepaid parking legend/map display
- refines Maroon & White Game popup field mappings for the updated service data
- renames the baseball event display name from `Baseball Parking` to `Baseball`

image
Recording 2026-04-06 111424

Recording 2026-04-06 111424

image

Closes #711  
Closes #710

### DIFF
--- a/libs/ts/events/ngx/src/lib/modules/builder/modules/accommodations/accommodations.component.ts
+++ b/libs/ts/events/ngx/src/lib/modules/builder/modules/accommodations/accommodations.component.ts
@@ -18,6 +18,7 @@ interface AccommodationChoiceGroup {
   styleUrls: ['./accommodations.component.scss', '../builder-module-base/builder-module-base.component.scss']
 })
 export class AccommodationsComponent implements OnInit {
+  public config = this.eventSettingsService.eventConfiguration()?.configuration;
   public savedOptionValue: Observable<string | boolean | number | null>;
 
   public accommodation$: Observable<SpecialEventOption>;
--- a/libs/ts/football/ngx/src/lib/modules/builder/builder.component.html
+++ b/libs/ts/football/ngx/src/lib/modules/builder/builder.component.html
@@ -1,7 +1,6 @@
 <div class="container12">
   <div class="flex flex-column">
     <h1 id="component-title">LONE STAR SHOWDOWN EVENT MAPS</h1>
-    <h4><a href="https://12thman.com/sports/football/schedule" target="_blank" rel="noopener noreferrer">Click here for Football dates.</a></h4>
   </div>
 
   <router-outlet></router-outlet>
--- a/libs/ts/football/ngx/src/lib/modules/builder/builder.component.html
+++ b/libs/ts/football/ngx/src/lib/modules/builder/builder.component.html
@@ -1,6 +1,7 @@
 <div class="container12">
   <div class="flex flex-column">
     <h1 id="component-title">LONE STAR SHOWDOWN EVENT MAPS</h1>
+    <h4><a href="https://12thman.com/sports/football/schedule" target="_blank" rel="noopener noreferrer">Click here for Football dates.</a></h4>
   </div>
 
   <router-outlet></router-outlet>
--- a/libs/ts/football/ngx/src/lib/modules/builder/builder.component.scss
+++ b/libs/ts/football/ngx/src/lib/modules/builder/builder.component.scss
@@ -3,10 +3,6 @@
 :host {
   background: #500000;
   color: #eeeeee;
-
-  h4 a {
-    color: #eeeeee;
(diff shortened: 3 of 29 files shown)


libs/ts/events/ngx/src/lib/modules/builder/modules/accommodations/accommodations.component.ts
```diff
@@ -18,6 +18,7 @@ interface AccommodationChoiceGroup {
   styleUrls: ['./accommodations.component.scss', '../builder-module-base/builder-module-base.component.scss']
 })
 export class AccommodationsComponent implements OnInit {
+  public config = this.eventSettingsService.eventConfiguration()?.configuration;
```
danielgoldbergtamu, line 21: I'm confused why this is needed now but was not before. Did this component not use a config before but needs to now?

libs/ts/football/ngx/src/lib/modules/builder/builder.component.html
```diff
@@ -1,6 +1,7 @@
 <div class="container12">
   <div class="flex flex-column">
     <h1 id="component-title">LONE STAR SHOWDOWN EVENT MAPS</h1>
+    <h4><a href="https://12thman.com/sports/football/schedule" target="_blank" rel="noopener noreferrer">Click here for Football dates.</a></h4>
```
danielgoldbergtamu: Why is this one hard-coded in the builder component? Just as an example that is overridden?

danielgoldbergtamu: Is it possible that we don't need this /ts/football/... component any longer if it can be added into the same place in the same way as all of the other config files?

libs/ts/football/ngx/src/lib/modules/builder/builder.component.scss
```diff
@@ -3,6 +3,10 @@
 :host {
   background: #500000;
   color: #eeeeee;
+
```
danielgoldbergtamu: Same. Is it possible that we don't need this /ts/football/... component any longer if it can be added into the same place in the same way as all of the other config files?

danielgoldbergtamu (changes requested):
A few things to check.

libs/ts/events/ngx/src/lib/modules/builder/modules/accommodations/accommodations.component.ts
sloneearly: This was added because of the new schedule link (`scheduleURL`) functionality. `config` gives the template access to the event's `EventConfiguration` object. So if a given event has a `scheduleUrl` set in its config file, the link will show on the builder page. If not, the `*ngIf` hides it. The `config?` name fills in the event's display name in the link text.

libs/ts/football/ngx/src/lib/modules/builder/builder.component.html
sloneearly: Fixed.

libs/ts/football/ngx/src/lib/modules/builder/builder.component.scss
sloneearly: Fixed.

danielgoldbergtamu (approved):
Thanks.

PR #743, opened 2026-04-20
feat(aggiema-angular): Add titles to maps header on sidebar

### LINKED ISSUE #738: aggiemap-angular: Move title of each map to the header
The title of each map should be moved to the header, below Aggie Map & search bar:

Image

sloneearly:
This PR adds the title of the map + the type of map to the side bar, right below the search bar and above the layers:

(Names are based on the `applicationName` in each map's definitions file)

image
image
image

Closes #738


danielgoldbergtamu (approved):
I don't love the font, or that it breaks the line, but this is fine for now.

PR #744, opened 2026-04-22
fix(aggiemap-angular): Update Sustainable Transporation layers to only pull from BikeMap server

### LINKED ISSUE #741: aggiemap-angular: Sustainable Transportation Map - stand-alone - fails to load
### Issue

1. It seems that the [standalone version of the sustainable transportation map](https://aggiemap.tamu.edu/parking/sustainable-transportation) may not have the correct layers set. 
2. It is not loading, which usually happens when the layers cannot be found. 
3. The layers on the main aggie map page load just fine.

### Resolution

1. Make sure that the standalone map is using the same layers as those on the main map

sloneearly:
This PR updates the Sustainable Transporation map (and the layers toggleable on the homepage) to pull only from https://gis.it.tamu.edu/arcgis/rest/services/TS/BikeMap/MapServer, rather than BikeMap and the EV Charge Stations server.

[
Screenshot 2026-04-22 082953
](url)

Closes #741


danielgoldbergtamu (approved):
Seems fine.

PR #757, opened 2026-04-24
fix(aggiemap-angular): Update Main Campus RNS and Line Paint layer visibility

### LINKED ISSUE #747: aggiemap-angular: Campus Main Parking Map missing layers
### Issue

Image

It should look like this:

Image

sloneearly:
This PR re-adds the fix for adding the RNS and Line Paint layers on the Main Campus Parking map.

image

Closes #747


danielgoldbergtamu (approved):
good.

PR #759, opened 2026-04-24
feat(aggiemap-angular): Re-added EV Charger layers to other maps, made standalone EV map

### LINKED ISSUE #752: aggiemap-angular: Create new EV Charger map
### Issue

1. Sustainable transportation needs a map with only EV Chargers on it.

This map just needs the EV Chargers layer (0) from the bike map rest services: https://gis.it.tamu.edu/arcgis/rest/services/TS/BikeMap/MapServer 

2. We still need EV Chargers on the bike map, so they will need to stay put and also live on their own, separate map.

sloneearly:
This PR re-adds the EV Charger layer to the Sustainable Transporation map and the main layers on the homepage. This PR also creates an independent EV Chargers map.

image

image

image

Closes #752

### DIFF
--- a/libs/aggiemap/ngx/common/src/lib/layer-sources/layer-sources.ts
+++ b/libs/aggiemap/ngx/common/src/lib/layer-sources/layer-sources.ts
@@ -22,6 +22,26 @@ export function LayerSources(
   options?: IFactoryExcludeOptions<IComposedIDefinitions>
 ): Array<LayerSource> {
   const bikeMapUrl = 'https://gis.it.tamu.edu/arcgis/rest/services/TS/BikeMap/MapServer';
+  const bikeMapEvChargeStationsLayerSource: LayerSource = {
+    type: 'feature',
+    id: 'ev-charge-stations-layer',
+    title: 'EV Charge Stations (Main + RELLIS)',
+    url: `${bikeMapUrl}/0`,
+    listMode: 'show',
+    visible: true,
+    popupComponent: Popups.MarkdownPopupComponent,
+    popupData: {
+      name: '{attributes.EV_ID}',
+      description:
+        '<strong>Network</strong>: {attributes.EV_Network}\n' +
+        '<strong>Charging Level</strong>: {attributes.Ch_Level}\n' +
+        '<strong>Garage Level</strong>: {attributes.Garage_Lvl}\n' +
+        '<strong>Notes</strong>: {attributes.EVCS_Notes}'
+    },
+    native: {
+      ...commonLayerProps
+    }
+  };
   const all: Array<LayerSource> = [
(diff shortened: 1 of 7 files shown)


libs/aggiemap/ngx/common/src/lib/layer-sources/layer-sources.ts
```diff
@@ -22,6 +22,26 @@ export function LayerSources(
   options?: IFactoryExcludeOptions<IComposedIDefinitions>
 ): Array<LayerSource> {
   const bikeMapUrl = 'https://gis.it.tamu.edu/arcgis/rest/services/TS/BikeMap/MapServer';
+  const bikeMapEvChargeStationsLayerSource: LayerSource = {
```
danielgoldbergtamu: @sloneearly Can this source definition be in the definition file for the layer instead of here in the layer-sources main file?

danielgoldbergtamu (changes requested):
Does this new feature definition have to be in this shared filed? Can it be placed into the definition of the layer instead?

danielgoldbergtamu (changes requested):
1. See if you can get rid of layersource on 25 and just have it in the body below.
2. See if you can have the layer's URL in the same place as all of the other ones.

libs/ts/events/ngx/src/lib/definitions/ev-chargers.definitions.ts
```diff
@@ -0,0 +1,75 @@
...
+  AggiemapCustomMapConfiguration,
+  EventConfiguration,
+  SpecialEventOptions
+} from '../interfaces/special-event.interface';
+
+export enum EV_CHARGERS_LAYERS {
+  EV_CHARGERS = 'ev-chargers'
+}
+
+const bikeLayersUrl = 'https://gis.it.tamu.edu/arcgis/rest/services/TS/BikeMap/MapServer';
```
danielgoldbergtamu: Get this from the connections file.

libs/ts/events/ngx/src/lib/definitions/sustainable-transportation.definitions.ts
```diff
@@ -18,6 +18,27 @@ export enum SUSTAINABLE_TRANSPORTATION_LAYERS {
 
 const bikeLayersUrl = 'https://gis.it.tamu.edu/arcgis/rest/services/TS/BikeMap/MapServer';
```
danielgoldbergtamu: Get this from the connections file

danielgoldbergtamu (changes requested):

danielgoldbergtamu (approved):
Will will pull the connection strings out as part of #768

PR #763, opened 2026-04-27
fix(aggiemap-angular): Fix POI Search by separating text queries from exact OBJECTID lookup

### LINKED ISSUE #750: aggiemap-angular: poi layers search source issues
### Issue

The POI layer is behaving strangely in the search box on the home page.

**Data Sources**

The POI data are defined in this [ArcGIS Online layer](https://tamu-fam.maps.arcgis.com/home/item.html?id=40d3d6fdacde4a43a8c7b99cbcac0696&dataTabView=table#data)

The search layer is in this [ArcGIS Online feature service](https://services1.arcgis.com/oxXAea6csqnDZ6WT/arcgis/rest/services/Points_of_Interest_view/FeatureServer/0//query?f=json&resultRecordCount=5&outFields=*&outSR=4326&returnGeometry=true&spatialRel=esriSpatialRelIntersects&where=UPPER(name)%20LIKE%20%27%251%25%27%20OR%20OBJECTID%20=%20%271%27).

**Examples**

1. Entering the term "1" finds "Bonfire Memorial" because it matches the ID=1 in the search feature layer

Image

2. Entering the term "Wright" does not find the new feature "Wright Gallery" added into the [AGOL Layer](https://tamu-fam.maps.arcgis.com/home/item.html?id=40d3d6fdacde4a43a8c7b99cbcac0696&dataTabView=table#data)

### Suspicions
1. The search source is not actually pulling from the data layer, it is pulling from the feature layer view, which has to be republished every time new data have been added. 

### Resolution
1. Is it possible to pull straight from the data source, not the feature layer?

sloneearly:
## Summary

This PR fixes POI search failures on the Aggie Map home page by separating free-text POI search from exact POI ID lookup.

Previously, the POI search source built a single `where` clause that searched both the POI name field and `OBJECTID`. That worked for numeric input such as `1`, but it caused ArcGIS to reject text queries such as `Wright` because the generated predicate attempted to compare a numeric `OBJECTID` field to a quoted text value. As a result, valid POIs like **Wright Gallery** were missing from search results even though they were present in the underlying feature service response.

This PR removes that invalid mixed query behavior while preserving existing POI deep-link/share URL behavior.

---

## Before: 

image

## After:

image

Search works for all POIs.

## Problem

The POI source was configured to search with both of these conditions:

- `name LIKE ...`
- `OBJECTID = ...`

For a text search such as `Wright`, the app generated a `where` clause effectively equivalent to:

```sql
UPPER(name) LIKE '%WRIGHT%' OR OBJECTID = 'WRIGHT'
```

That second predicate is invalid because:

- `OBJECTID` is numeric
- the shared query builder inserts the search term as a quoted string

ArcGIS rejected the entire query with:

```json
{
  "error": {
    "code": 400,
    "message": "Cannot perform query. Invalid query parameters."
  }
}
```

This is why:
- searching `1` could return **Bonfire Memorial** via `OBJECTID`
- searching `Wright` could fail to return **Wright Gallery**

---

## Solution

The fix splits POI handling into two separate search sources:

### 1. `points-of-interest`
This remains the visible, user-facing POI search source.

It now performs **text-only** matching against the POI name field:

- `name LIKE '%term%'`

This avoids sending invalid `OBJECTID` comparisons for text input.

### 2. `points-of-interest-exact`
This is a hidden source used only for exact POI lookup by ID.

It performs:

- `OBJECTID = <id>`

This preserves existing deep-link/share-link behavior such as:

```text
?poi=123
```

---

## Why this approach

This change addresses the actual failure mode reported in the issue: the generated POI query was invalid.

It does **not** switch Aggie Map from the current public POI view service to a direct source layer. During investigation, the current public POI endpoint was confirmed to contain the expected data, including the relevant example, which means the immediate bug was query construction rather than missing data in the current service.

Additionally, the likely direct source service appears to require authentication, while the current POI view is publicly queryable. Repointing to the direct source would therefore be a separate change with additional auth implications, not part of the search-query bug fix.

---

### `libs/aggiemap/ngx/common/src/lib/search-sources/search-sources.ts`

- Added hidden `points-of-interest-exact` source for exact `OBJECTID` lookup
- Updated visible `points-of-interest` source to search only by `name`
- Added the new source to `ComposedSearchSourcesKeyMap`

### `libs/aggiemap/ngx/popups/src/lib/components/poi/poi.component.ts`

- Updated POI share-link generation to use `points-of-interest-exact`
- Preserves `?poi=<OBJECTID>` deep links

### `libs/aggiemap/ngx/common/src/lib/search-sources/search-sources.spec.ts`

- Updated source-count expectations
- Added regression coverage ensuring:
  - text POI search uses only `name`
  - exact POI lookup remains available through the hidden source
  - deep-link parameter config remains intact

---

Closes #750

### DIFF
--- a/libs/aggiemap/ngx/common/src/lib/search-sources/search-sources.spec.ts
+++ b/libs/aggiemap/ngx/common/src/lib/search-sources/search-sources.spec.ts
@@ -1,7 +1,16 @@
-import { ComposedSearchSourcesKeyMap, SearchSources } from './search-sources';
 import { IComposedConnections } from '../connections';
 import { IComposedIDefinitions } from '../definitions';
 import { IFactoryExcludeOptions } from '../utils/definitionFactory';
+
+jest.mock('@tamu-gisc/aggiemap/ngx/popups', () => ({
+  Popups: {
+    BuildingPopupComponent: 'BuildingPopupComponent',
+    ParkingLotPopupComponent: 'ParkingLotPopupComponent',
+    PoiPopupComponent: 'PoiPopupComponent'
+  }
+}));
+
+import { ComposedSearchSourcesKeyMap, SearchSources } from './search-sources';
 
 describe('SearchSources', () => {
   const connections: IComposedConnections = {
@@ -12,7 +21,9 @@ describe('SearchSources', () => {
     bikeLocationsUrl: 'https://example.com/bike-locations',
     constructionUrl: 'https://example.com/construction',
     inforUrl: 'https://example.com/infor',
-    tsMainUrl: 'https://example.com/ts-main'
+    tsMainUrl: 'https://example.com/ts-main',
+    poiUrl: 'https://example.com/poi-service',
+    diningLocationsUrl: 'https://example.com/dining-locations'
   };
 
   const definitions: IComposedIDefinitions = {
@@ -27,12 +38,6 @@ describe('SearchSources', () => {
       layerId: 'poi-layer',
       name: 'Points of Interest',
       url: 'https://example.com/poi'
-    },
-    RESTROOMS: {
-      id: 'restrooms',
-      layerId: 'restrooms-layer',
-      name: 'Restrooms',
-      url: 'https://example.com/restrooms'
     },
     BUILDINGS: {
       id: 'buildings',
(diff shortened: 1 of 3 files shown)


danielgoldbergtamu (approved):
Nice.

PR #767, opened 2026-04-29
fix(aggiemap-angular): Replace RNS labels with actual space ID

### LINKED ISSUE #747: aggiemap-angular: Campus Main Parking Map missing layers

sloneearly:
This PR replaces the RNS labels with their actual `RNS_Num` instead of just "RNS."

image

Closes #747


danielgoldbergtamu (approved):
looks good.

PR #773, opened 2026-05-01
feat(aggiemap-angular): Add Requesting Maps & Changes info page

sloneearly:
Pre-requisite: #774 

This PR adds the new page "Requesting New Aggie Maps & Changes to Existing Maps" to AggieMap and adds a "Requesting Maps & Changes" link has been added to the help menu (question mark icon) on the main map page.

image
image
image

Closes #770

PR #772, opened 2026-05-01
feat(aggiemap-angular): Replace Aggie Map privacy page with link to TAMU's privacy page

sloneearly:
Pre-requisite: #774 

This PR removes the custom privacy page entirely and updates all "Privacy & Security" links (help menu, footer, and mobile sidebar) to point directly to the official TAMU privacy statement at tamu.edu, https://www.tamu.edu/statements/privacy.html.

https://github.com/user-attachments/assets/9b2feea3-aad6-4da8-9735-8c3ba27c920b

Closes #771

PR #775, opened 2026-05-01
fix(aggiemap-angular): Remove forbidden static import of lazy-loaded

### LINKED ISSUE #774: aggiemap-angular: aggiemap-ngx-core lint fails with "Imports of lazy-loaded libraries are forbidden" when affected by unrelated changes
The `aggiemap-ngx-core` library has a pre-existing lint error that goes undetected in CI until a PR happens to touch any file inside that library, at which point the linter runs on the whole project and surfaces the violation.

## Error example

> nx run aggiemap-ngx-core:lint
> 
> /libs/aggiemap/ngx/core/src/lib/services/event-notifications/event-notifications.service.ts
> Error: 3:1  error  Imports of lazy-loaded libraries are forbidden  @nx/enforce-module-boundaries

`aggiemap-ngx-core` already lazy-loads that same library via the Angular router:
`{ path: 'events', loadChildren: () => import('@tamu-gisc/ts/events/ngx').then((m) => m.TsEventsNgxModule) }
`

Because `aggiemap-ngx-core` is the library responsible for lazy-loading `ts-events-ngx`, `NX's @nx/enforce-module-boundaries` rule classifies `ts-events-ngx` as a lazy-loaded dependency of that project and forbids any static imports of it from within the same project.

## Resolution

Replace the static import with an 'InjectionToken'. The service defines a minimal local interface and accepts the definitions as an optional injected dependency. The actual 'EventDefinitions' array is provided in 'app.module.ts' inside the aggiemap-angular app, which is an application entry point and is not subject to the same module boundary restrictions as library code.

sloneearly:
This PR replaces direct EventDefinitions import in EventNotificationsService with an InjectionToken, providing the definitions from app.module.ts where the ts-events-ngx import is permitted.

`event-notifications.service.ts` was statically importing `EventDefinitions` directly from `@tamu-gisc/ts/events/ngx.` Because `aggiemap-ngx-core` already lazy-loads that same library via the Angular router (`loadChildren`), NX's `@nx/enforce-module-boundaries` rule treats it as a lazy-loaded library and forbids any static imports of it from within the same project, causing a lint error in CI whenever `aggiemap-ngx-core` appears in the affected set.

The fix replaces the static import with an `InjectionToken (EVENT_NOTIFICATION_DEFINITIONS)`. The service now declares a minimal local interface for the data it needs and injects the token as an optional dependency. The actual `EventDefinitions` array is provided in `app.module.ts` inside the aggiemap-angular app, which is an application entry point and is permitted to import from `@tamu-gisc/ts/events/ngx` without restriction. Runtime behavior is unchanged.

Required for #772 and #773
Closes #774

### DIFF
--- a/apps/aggiemap-angular/src/app/app.module.ts
+++ b/apps/aggiemap-angular/src/app/app.module.ts
@@ -6,8 +6,9 @@ import { HttpClientModule } from '@angular/common/http';
 import { Angulartics2Module } from 'angulartics2';
 
 import { env, EnvironmentModule } from '@tamu-gisc/common/ngx/environment';
-import { AggiemapNgxCoreModule } from '@tamu-gisc/aggiemap/ngx/core';
+import { AggiemapNgxCoreModule, EVENT_NOTIFICATION_DEFINITIONS } from '@tamu-gisc/aggiemap/ngx/core';
 import { NotificationModule, notificationStorage } from '@tamu-gisc/common/ngx/ui/notification';
+import { EventDefinitions } from '@tamu-gisc/ts/events/ngx';
 
 import * as environment from '../environments/environment';
 import { AppComponent } from './app.component';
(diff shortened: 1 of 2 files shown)


danielgoldbergtamu (changes requested):
Seems ok, but can you make it so that it does not remove the comments?

sloneearly:
> Seems ok, but can you make it so that it does not remove the comments?

Comments have been reapplied.

danielgoldbergtamu (approved):
Thanks.

PR #776, opened 2026-05-01
fix(aggiemap-angular): Dismiss search results on Escape and click-outside

### LINKED ISSUE #769: aggiemap-angular: search dropdown stuck - no escape option
### Issue

When a user types into the search box on the main Aggie Map and a list of results shows, there is no way to get out of the search.

Escape would be the normal way to dismiss the search results popup, but it does not work. 

### Resolution

Dismiss the search box results when

1. Escape is pressed.
2. The user clicks away from the search bar or search box i.e., clicks on anything outside of the search results. 
3. The users clicks back into the search bar.

sloneearly:
This PR makes it where search results now close when the user presses Escape or clicks anywhere outside the search component. The mobile input also gains the missing blur handler so clicking away behaves consistently with desktop.

https://github.com/user-attachments/assets/09370648-18c2-49e4-89d8-9f9af7694fdc

Closes #769

### DIFF
--- a/libs/ui-kits/ngx/search/src/lib/components/search-mobile/search-mobile.component.html
+++ b/libs/ui-kits/ngx/search/src/lib/components/search-mobile/search-mobile.component.html
@@ -9,7 +9,7 @@
       <div></div>
     </div>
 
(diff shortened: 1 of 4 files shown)


danielgoldbergtamu (approved):
Nice.

PR #777, opened 2026-05-01
feat(aggiemap-angular): Centralize shared layer source URLs for event maps

### LINKED ISSUE #768: aggiemap-angular: Pull layer sources urls from connections file files rather than hard coding in definitions files
### Issue
There are at least two places where layer source URLs are hard coded into definitions files rather than being set in the connections files.

This is not ideal because then we do not have a way to automatically point the dev sites to dev data layers.

1. libs/aggiemap/ngx/common/src/lib/definitions.ts

```    
      name: 'AggiePrint Locations',
      url: 'https://services1.arcgis.com/qr14biwnHA6Vis6l/ArcGIS/rest/services/TAMUPrinters/FeatureServer/0', 
```

2. libs/aggiemap/ngx/common/src/lib/definitions.ts

``` 
      name: 'Single Occupancy Restroom Locations',
      url: 'https://gis.tamu.edu/arcgis/rest/services/FCOR/MapInfo_20190529/MapServer/1', 
```

### Resolution

Place the URL connections for these layers into the connections file and use the connection variables to set them in the definitions.

sloneearly:
This PR cleans up how AggieMap layer source URLs are managed by moving map URLs into the shared 'connections.ts config file and updating event map definitions to reference those shared values instead of hardcoded URLs.

It also updates the previously hardcoded common layer definitions, including AggiePrint and single-occupancy restrooms, so source configuration is more consistent across the app. 

Closes #768

### DIFF
--- a/libs/aggiemap/ngx/common/src/lib/definitions.ts
+++ b/libs/aggiemap/ngx/common/src/lib/definitions.ts
@@ -96,14 +96,14 @@ export function Definitions(Connections: IComposedConnections): IComposedIDefini
       id: 'aggieprint-locations',
       layerId: 'aggieprint-locations-layer',
       name: 'AggiePrint Locations',
-      url: 'https://services1.arcgis.com/qr14biwnHA6Vis6l/ArcGIS/rest/services/TAMUPrinters/FeatureServer/0',
+      url: `${Connections.aggiePrintUrl}/0`,
       popupComponent: Popups.MarkdownWDirectionsPopupComponent
     },
     SINGLE_OCCUPANCY_RESTROOMS: {
       id: 'single-occupancy-restroom-locations',
       layerId: 'single-occupancy-restroom-locations-layer',
       name: 'Single Occupancy Restroom Locations',
-      url: 'https://gis.tamu.edu/arcgis/rest/services/FCOR/MapInfo_20190529/MapServer/1',
+      url: `${Connections.inforUrl}/1`,
       popupComponent: Popups.MarkdownWDirectionsPopupComponent
     },
     BIKE_DISMOUNT_ZONES: {
--- a/libs/ts/events/ngx/src/lib/definitions/4h-roundup.definitions.ts
+++ b/libs/ts/events/ngx/src/lib/definitions/4h-roundup.definitions.ts
@@ -15,7 +15,7 @@ export enum FOUR_H_ROUNDUP_LAYERS {
   FOUR_H_ROUNDUP_LOCATIONS = 'four-h-roundup-locations'
 }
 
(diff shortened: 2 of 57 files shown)


libs/aggiemap/ngx/common/src/lib/definitions.ts
```diff
@@ -96,14 +96,14 @@ export function Definitions(Connections: IComposedConnections): IComposedIDefini
...
-      url: 'https://services1.arcgis.com/qr14biwnHA6Vis6l/ArcGIS/rest/services/TAMUPrinters/FeatureServer/0',
+      url: `${Connections.aggiePrintUrl}/0`,
       popupComponent: Popups.MarkdownWDirectionsPopupComponent
     },
     SINGLE_OCCUPANCY_RESTROOMS: {
       id: 'single-occupancy-restroom-locations',
       layerId: 'single-occupancy-restroom-locations-layer',
       name: 'Single Occupancy Restroom Locations',
-      url: 'https://gis.tamu.edu/arcgis/rest/services/FCOR/MapInfo_20190529/MapServer/1',
+      url: `${Connections.inforUrl}/1`,
```
danielgoldbergtamu, line 106: This is the way that we want to do it since it calls the connection by name which is filled in with the $gishost (dev/prod) in the connections file.

```
url: `${Connections.NAME}/NUMBER`,

```

libs/ts/events/ngx/src/lib/definitions/4h-roundup.definitions.ts
```diff
@@ -14,7 +15,7 @@ export enum FOUR_H_ROUNDUP_LAYERS {
   FOUR_H_ROUNDUP_LOCATIONS = 'four-h-roundup-locations'
 }
 
-const eventUrl = 'https://gis.it.tamu.edu/arcgis/rest/services/TS/4H_Roundup/MapServer';
+const eventUrl = Connections('gis.it.tamu.edu').fourHRoundupUrl;
```
danielgoldbergtamu: This is not the way that we want to do it since it hard codes the host URL (gis.tamu.edu) by name which is does not fill it in with the $gishost (dev/prod) in the connections file so that we can automatically have the layers switch to dev/prod based on the host it's running on.

```
const eventUrl = Connections('gis.it.tamu.edu').fourHRoundupUrl;

```

danielgoldbergtamu (changes requested):

libs/ts/events/ngx/src/lib/definitions/4h-roundup.definitions.ts
sloneearly: Just refactored it.

danielgoldbergtamu (approved):
Better, thanks!

PR #794, opened 2026-05-11
fix(aggiemap-angular) Align numbered list text on Requesting Maps page

### LINKED ISSUE #783: aggiemap-angular: requesting maps page alignment
### Issue

Formatting of list looks weird since it is <- of everything else

Image

### Resolution

Update CSS to move lists in line with the left alignment of the rest of the text.

sloneearly:
This PR aligns the numbered list at the bottom of the Requesting Maps page with the rest of the text:

image

Closes #783


danielgoldbergtamu (changes requested):
We may need to use this same formatting in other locations on different pages. It would better if you did not name the CSS classes specifically for this one page. Please make them general names instead so it's not super weird to re-use them on a different page with a different name in the future.

sloneearly:
> We may need to use this same formatting in other locations on different pages. It would better if you did not name the CSS classes specifically for this one page. Please make them general names instead so it's not super weird to re-use them on a different page with a different name in the future.

I went ahead and renamed it to `numbered-list` and put it in the global `styles.scss` since I figure it will be used semi-frequently like you said.

danielgoldbergtamu (approved):
Thank you!

Issue #793, opened 2026-05-11
aggiemap-angular: Maps with grouped layers showing completely random layers every time page is loaded


sloneearly:
On maps with grouped layers, such as:
- [Savannah Bananas baseball map](https://aggiemap.tamu.edu/events/savannah-bananas-parking/map/d)
- [Physics Festival](https://aggiemap.tamu.edu/events/phys-eng-fest/map/d)
- [Maroon and White Game](https://aggiemap.tamu.edu/events/maroon-white-game-2026/map/d)
- New High School graduation  (not published yet)

The map layers seem to be loading randomly every time the page is loaded or refreshed.

Banana map video: https://imgur.com/LX8WQAC

Physics festival video: https://imgur.com/JhpffEK

Maroon and White game video: https://imgur.com/6ZuE2Qv

High School Graduation video: https://imgur.com/67m9ZM5

Full album of issue videos: https://imgur.com/a/zADnufj

Clarifications:
- This is not an issue with zoom level or not waiting long enough for features to load
    - I tested every zoom level and the result is the same, no new layers appear
    - I have waited on each page for 5+ minutes and no new layers loaded in
- This is all on prod, except for the High School Graduation map that is on my localhost
- These are the only maps affected that I have found, so it is not just an issue with any map that has a dropdown (i.e Baseball map is fine)
- Not an issue exclusively with maps that use `type: 'group'` as the Maroon and White map does not use this type

danielgoldbergtamu:
I believe this may be a load balancing issue since the results change every other time you load the page. 50% of the time you get one set of results, 50% of the time you get the other.

Is it possible that the services are not in the same stop/start state across the two load balanced web adapters or GIS servers?

danielgoldbergtamu:
I am going to close this because it is a service issue, not an application issue. Unfortunately we will have to keep stopping starting the services when we see this behavior until we can get a resolution from Esri as to why the services keep crapping out.

PR #795, opened 2026-05-12
fix(aggiemap-angular) Fix popups for grouped layer maps

sloneearly:
This PR fixes an issue where popups were not loading on special event maps with grouped dropdown layers, like the Banana Ball Map and Physics Festival Map (when those layers appear, that is...):

image

image

This does not resolve issue #793 and is still being actively investigated.


sloneearly:
For clarification, I understand we can't push any new PRs until the #793 gets fixed, but knowing that popups are a separate issue will hopefully isolate the issue further.

danielgoldbergtamu (approved):
Let's give it a shot.

PR #800, opened 2026-05-13
feat(aggiemap-angular) 4H Roundup event map update

### LINKED ISSUE #790: aggiemap-angular: Update 4H Roundup Map
### Issue
1. The 4H Roundup map has been updated to now include icons on top of parking lots, which means nothing is currently showing up on the map.  
Map should look like this, with event location icons on top and event parking as the 2nd layer:

Image

2. The hyperlink needs to be updated, currently it is the link for 2025, I wasn't sure if we were doing dates in the link or removing them, but we need to get that updated so we can alert the event organization and it is updated on the 4H website and the TS website. https://aggiemap.tamu.edu/events/4h-roundup-2025/map/d 

*Dates for this event are June 1-4th, so it's not critical at this moment, but we will need it in the next few weeks.

sloneearly:
This PR updates the 4H Roundup map to use newly added features and update the date to be June 1-4, 2026.

image

Closes #790

### DIFF
--- a/libs/ts/events/ngx/src/lib/definitions/4h-roundup.definitions.ts
+++ b/libs/ts/events/ngx/src/lib/definitions/4h-roundup.definitions.ts
@@ -11,45 +11,30 @@ import {
 } from '../interfaces/special-event.interface';
 
 export enum FOUR_H_ROUNDUP_LAYERS {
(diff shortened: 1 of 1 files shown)


danielgoldbergtamu (approved):
Thanks.

PR #803, opened 2026-05-20
feat(aggiemap-angular): Refactor main homepage map to be structured same as other maps

### LINKED ISSUE #765: aggiemap-angular: generalize the main map to have a definitions file like all other sub-maps
### Issue

The main map is treated differently than all of the other maps in that it has its layers hard coded into the layer source rather than pulling from a definitions file: libs/aggiemap/ngx/common/src/lib/layer-sources/layer-sources.ts

### Resolution

Update the main map to use the same structure as all of the other maps by creating a definitions file for it so it has the same structure as everything else.

sloneearly:
Previously, the homepage (main) map was the only map without a dedicated definitions file. All of its layers were hardcoded in `layer.sources.` This PR extracts everything specific to the main map into a single dedicated file, `libs/aggiemap/ngx/common/src/lib/definitions/main.definitions.ts`, following the same structure already used by every other map. 

Close #765


danielgoldbergtamu (approved):
Thanks.

PR #805, opened 2026-05-22
fix(aggiemap-angular): Gracefully handle failed search sources

### LINKED ISSUE #791: aggiemap-angular: handle search source failures gracefully
### Issue

1. There were some underlying data set changes to the basemap data layers in the SDE which caused several of the basemap layers to a) cease to exist, and b) stop returning data.
2. Because of this, the front end application was receiving unexpected responses from the search queries which were not handled gracefully, meaning that the whole search stopped working even though other search sources were still functioning properly.
3. The search framework should be able to continue to work even if one or more of the sources are missing, corrupted, or otherwise returning invalid/unexpected data.

### Resolution
1. Build out error handling in the search pipeline to gracefully deal with one or more search sources returning no, incorrect, or otherwise unexpected data.

**Example uncaught errors which crash the search pipeline**

1. The layer was deleted

Image

sloneearly:
This PR makes the search pipeline resilient to any single source returning no data, an error envelope, or an unexpected payload. Failures are isolated to the offending source and the rest of the search continues to render.

## Changes
- All changes are in `libs/ui-kits/ngx/search/src/lib/services/search.service.ts.`
    - New `_safeRequest` helper. Wraps each per-source http.get with two guards: `catchError` converts transport-level errors (network failure, non-2xx status) to an empty response.
    -  A `map` step detects ArcGIS error envelopes `({ error: { code, message, details } })` returned with HTTP 200. Notably the "Layer not found" response when a referenced layer has been removed from the service and normalizes them to an empty response.
    - Both branches log a labelled `console.warn` so the failing source is still discoverable from DevTools.
  - Base and scoring requests now route through `_safeRequest.` Per-source failures no longer reject the outer `forkJoin`, so one bad source can't take down the whole pipeline.
  - Scoring merge is now defensive. Previously `[...responses.scoring[i][featuresLocation]`, ...`r[featuresLocation]]` would throw `TypeError: undefined` is not iterable when either response was an error envelope. Both arrays now fall back to `[]` when missing.
  - Final result mapping tolerates empty responses. An `empty-object` sentinel from `_safeRequest` cleanly produces features: `[]` for that source while siblings render normally.
   - Subscriber error backstop no longer re-throws. The previous `throw new Error(err)` tore down the active subscription and left consumers without a `SearchResult` to render. It now logs and emits an empty `SearchResult` on `_store`, so consumers always receive a usable value and the `_searching` indicator is reliably reset.

## Test with deleted layer (deleted basemap layer 1):

Screenshot 2026-05-22 095834

> Closes #791

### DIFF
--- a/libs/ui-kits/ngx/search/src/lib/services/search.service.ts
+++ b/libs/ui-kits/ngx/search/src/lib/services/search.service.ts
@@ -302,9 +302,10 @@ export class SearchService {
    * Downstream feature extraction (`r[featuresLocation] ? ... : []`) treats the empty object as a
    * no-results response for that source.
    */
-  private _safeRequest(url: string, sourceLabel: string): Observable<unknown> {
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  private _safeRequest(url: string, sourceLabel: string): Observable<Object> {
     return this.http.get(url).pipe(
-      map((response: unknown) => {
+      map((response) => {
         if (response && typeof response === 'object' && (response as { error?: unknown }).error) {
           console.warn(
             `Search source "${sourceLabel}" returned an error envelope:`,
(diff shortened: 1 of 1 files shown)


danielgoldbergtamu (approved):
Thanks.

PR #810, opened 2026-05-26
feat(aggiemap-angular) Custom width/height adjusting for icons on Legends and Maps

### LINKED ISSUE #722: aggiemap-angular: Create layer-level option for handling stretched icons on AggieMap
### Issue

There has recently been a repeating issue with icons on newer maps being stretched and/or visually inconsistent. The current solution has been to hardcode the icons into the definitions files, but this is obviously not sustainable or scalable. 

### Examples of stretching

- ## SEC Grounds 

Image

- ## Maroon & White 

Image

- ## Baseball
  
Image

### Proposed solution

Add an optional, layer-level override to `LayerSource`

Why:
- The problem is reoccuring layer issue
- The renderer already lives on `native.renderer`, so this stays close to the existing definition structure
- We already have event-level legend options for other behaviors, so this fits the repo’s current config style

Suggested behavior:
- By default, keep current behavior
- Allow a layer to opt-in and tell AggieMap to render its specfic icons differently
- Implement customizable parameters for icons, i.e: 
  - `legend.mode: 'arcgis' | 'renderer-symbol' | 'custom-src'`
  - `legend.preserveAspectRatio: true`
  - `legend.fit: 'contain'`
  - `legend.width`
  - `legend.height`

sloneearly:
There has been an ongoing issue with some map icons getting stretched when brought into AggieMap. This PR adds an opt-in, reusuable system for manually adjusting the width, height, and style of icons.

4H Map Icons before:

Screenshot 2026-05-26 105845

After:

image

This PR only adjusts the 4H map as an example. This can be applied to other maps in the future.

Closes #722

### DIFF
--- a/libs/common/types/src/lib/common-types.ts
+++ b/libs/common/types/src/lib/common-types.ts
@@ -388,7 +388,61 @@ export type LayerSource = LayerSourceType & {
    * protected resources.
    */
   auth?: LayerSourceAuthInfo;
+
+  /**
+   * Optional opt-in override that controls how the layer's icons are rendered in the legend.
+   *
+   * Without this property, the legend keeps its existing default behavior (the ArcGIS-generated
+   * preview is used as-is). Set this when the service-provided icon swatch is visually inconsistent
+   * (for example, stretched non-square picture markers) and the layer needs targeted control over
+   * its legend icons.
+   */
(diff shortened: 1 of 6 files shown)


danielgoldbergtamu (approved):
Nice.

PR #818, opened 2026-05-28
fix(aggiemap-angular) Fix URL for Break/Summer map

sloneearly:
This PR changes the Break/Summer URL from an "event" to "parking."

Closes ##816


danielgoldbergtamu (approved):
Looks good.

PR #826, opened 2026-06-01
fix(aggiemap-angular) Make visitor layer source same for homepage and main campus parking map

### LINKED ISSUE #825: aggiemap-angular: Wrong layer source for visitor parking on homepage map
### Issue

The visitor parking layer (along with many more) is broken on the dev version of the homepage map:

Image

However, it works just fine on the Main Campus Parking map:

Image

### Resolution

Change the layer source for the homepage map to be the same as the main campus parking map.

sloneearly:
This PR makes the visitor parking layer on the homepage map (not working) pull from the same source as the main campus parking map (working.)

image

Closes #825


danielgoldbergtamu (approved):
Thanks.

PR #832, opened 2026-06-02
fix(aggiemap-angular): Fix map canvas going blank after repeated 2D/3D toggling

### LINKED ISSUE #556: aggiemap-angular: 2d to 3d to 2d to 3d workflow/clicks nukes map
### Issue 

If the user undertakes a specific set of 2D to 3D interface steps, the entirety of the map canvas goes blank.

### Steps to reproduce

1. Open [Aggie Map](https://aggiemap.tamu.edu) on desktop. It will start in 2D.
2. Click on 3D to transition to the 3D view.
3. Click on 2D to transition back to the 2D view.
4. Click on 3D again.
5. Map canvas is nuked.

Image

sloneearly:
This PR fixes an issue where toggling the map perspective in a 2D → 3D → 2D → 3D sequence nuked the
entire map canvas. The second switch to 3D rendered nothing but the empty background grid.

### Changes

Reworked `setView` to:

- Give every view its own fresh container element (discarding the previous one).
- Disconnect the outgoing view from the shared map and reconnect the incoming
  view, so only the active view processes layers.
- Capture the current viewpoint and restore it onto the incoming view so the
  camera/extent carries across the swap.

No `view.destroy()` is used, since destroying a view in the ArcGIS JS API also
destroys the shared map and layers.

### Map after 4th 3D toggle:
image

Closes #556

### DIFF
--- a/libs/maps/esri/src/lib/services/map/map.service.ts
+++ b/libs/maps/esri/src/lib/services/map/map.service.ts
@@ -159,11 +159,48 @@ export class EsriMapService {
   public setView(view: esri.SceneView | esri.MapView) {
     this.destroyViewClickEventHandler();
 
-    this._modules.view.container = null;
-    this._modules.view = null;
-    this._modules.view = view;
+    // Carry the current camera/extent across the swap. A reused view otherwise re-attaches with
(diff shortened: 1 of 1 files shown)


danielgoldbergtamu (approved):
Nice.

PR #833, opened 2026-06-02
feat(aggiemap-angular) Auto pan 3D toggle

### LINKED ISSUE #554: aggiemap-angular: 3D auto-pan angle
### Issue

When 3D is clicked on Aggie Map, users cannot tell that 3D buildings have appeared because the map is looking at the top down.

Image

### Proposed Solution

When the 3D option is selected on Aggie Map, the map should auto-pan to an angle different than top down (45 degrees or whatever works) the 3D buildings load so highlight that the map is indeed in 3D and there are 3D buildings to explore and marvel at.

Image

sloneearly:
This PR automatically pans the 3D mode to 45 degrees when toggled, showing that the buildings are actually in 3D and not just a flat map.

image

Closes #554


danielgoldbergtamu (approved):
Looks good.

PR #839, opened 2026-06-05
fix(aggiemap-angular): Allow links to buildings to use abbreviations and number code

### LINKED ISSUE #838: aggiemap-angular: allow incoming links to buildings to include the abbreviation in addition to the building code
### Issue

1. The registrar uses links within the student scheduling app that sends users to Aggie Map via a URL that includes the building abbreviation in the query string as follows: [https://aggiemap.tamu.edu/map/d?ref=howdy&BldgAbbrv=WCBA](https://aggiemap.tamu.edu/map/d?ref=howdy&BldgAbbrv=WCBA)
2. Sometime in the last few cycles, this stopped working, allowing instead only links that include the building code: [https://aggiemap.tamu.edu/?bldg=1510](https://aggiemap.tamu.edu/?bldg=1510)
3. We need to support both the building code and the building abbreviation

### Resolution

1. Update the incoming query parameter processing to get the buildings layer to support selection by building abbreviation again.

sloneearly:
This PR fixes an issue where using an incoming link to AggieMap with a building abbreviation did not work:

image

URLs from Howdy work as well:

image

URLs with the building number code also still work:

image

Closes #838

### DIFF
--- a/libs/aggiemap/ngx/common/src/lib/search-sources/search-sources.spec.ts
+++ b/libs/aggiemap/ngx/common/src/lib/search-sources/search-sources.spec.ts
@@ -13,7 +13,9 @@ jest.mock('@tamu-gisc/aggiemap/ngx/popups', () => ({
 import { ComposedSearchSourcesKeyMap, SearchSources } from './search-sources';
 
 describe('SearchSources', () => {
-  const connections: IComposedConnections = {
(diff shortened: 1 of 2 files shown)


danielgoldbergtamu (approved):
Thanks!

PR #860, opened 2026-06-30
feat(aggiemap-angular) Move-In map overhaul

### LINKED ISSUE #848: aggiemap-angular: Move In Map Updates
Move in begins 8/18, need map up and functioning by 8/3

### Issue 1: Options shown by date

Have the specific parking arrangements show by date for Move-In Streets layer and Move-In Lots layer. These layers both have start and end dates/times. Move-In Points of Interest needs to be on by default

### Issue 2: Create 3 zooms/maps for campus areas

Using the CampusArea field update the zoom to the area on campus: Northside, Southside and White Creek. Once through the builder the section of campus where the residence hall is located should be the only area shown

### Issue 3: Layers off by default

Parking lot closures and icons will be toggled off by default

### Issue 4: Split out icons into separate layers

Split the icons into separate layers so they can be toggled on/off independently

### Issue 5: Rename accessible options

Rename Yes/No options to: "No, I do not need accessible parking" and "Yes, I do require accessible parking"

### Issue 6:

Have accessible lots/parking show only on the accessible parking needed option, remove them from the no I don't need accessible parking map

### Issue 7: Identify the residence hall on the map

Once through the builder the residence hall selected should flash/identify so users know its location

sloneearly:
This PR overhauls the Move-In builder and map, making it work as intended as resolves all the issues listed in #848.

Issue 1: Show parking options by date
The Move-In Streets and Move-In Lots layers now change based on the day picked in the builder. Each arrangement has a start and end date in the data, so the map only shows the ones in effect on the selected day. Points of Interest are not tied to a date.

image

Issue 2: Zoom to the campus area (Northside / Southside / White Creek)
Every residence hall is tagged with its campus area. After the builder, the map zooms in close to that area and hides the features in the other two areas, so you only see the section of campus where your hall is. Campus-wide visitor lots (general Free/Paid parking) that aren't tied to one area stay visible.

image

Issue 3: Closures and icons off by default
Parking lot closures, the street "No Roadside Parking" closures, and all the icon layers now start turned off. The user turns on whatever they want from the layer list.

image
image

Issue 4: Split icons into separate layers
The single Points of Interest layer was split into one layer per icon: Check-In/Key Pickup, Information, Dining, Cardboard Recycling, Bike Registration, and No Roadside Parking are now grouped under a "Points of Interest" heading so each can be toggled on and off on its own. The Move-In Streets and Move-In Lots layers were split the same way (drop-off zones vs. closures; available vs. accessible vs. closures).

image

Issue 5: Rename the accessible parking options
The Yes/No choices are now labeled "No, I do not need accessible parking" and "Yes, I do require accessible parking."

image

Issue 6: Accessible parking only on the "Yes" map
Accessible parking is its own layer that's hidden by default and only appears when the user picks "Yes, I do require accessible parking." 

image
image

Issue 7: Identify the residence hall
The selected hall is filled in bright cyan and briefly flashes when the map loads so users can spot it right away. (This also required matching halls on the basemap's Number field, which works on both the production and dev servers.)

image

https://github.com/user-attachments/assets/193c0d4a-2672-4115-82e8-3ad503d13fb8

Issue 8: Fix stretched pin icons
A few icons (Bike Registration, Check-In/Key Pickup, No Roadside Parking) were being squished into a square. They now render at their correct tall pin proportion (24×30) so they no longer look stretched.

Closes #848

### DIFF
--- a/libs/ts/events/ngx/src/lib/definitions/move-in.definitions.ts
+++ b/libs/ts/events/ngx/src/lib/definitions/move-in.definitions.ts
@@ -11,54 +11,176 @@ import {
 } from '../interfaces/special-event.interface';
 
 export enum MOVE_IN_LAYERS {
-  MOVE_IN_POI = 'Move-In Points of Interest',
-  MOVE_IN_STREETS = 'Move-In Streets',
-  MOVE_IN_LOTS = 'Move-In Lots',
-  RESIDENCE_HALL = 'Residence Hall'
+  RESIDENCE_HALL = 'Residence Hall',
+  // Points of Interest is split into one feature layer per icon Type so each can be toggled
+  // independently (grouped under a single parent in the layer list).
+  POI_GROUP = 'Move-In Points of Interest',
+  POI_KEYS = 'move-in-poi-keys',
+  POI_INFO = 'move-in-poi-info',
+  POI_DINING = 'move-in-poi-dining',
+  POI_RECYCLE = 'move-in-poi-recycle',
+  POI_BIKE = 'move-in-poi-bike',
+  POI_NO_PARKING = 'move-in-poi-no-parking',
+  // Streets are split so the No-Roadside-Parking closures can be toggled off independently of the
+  // drop-off zones (grouped under a single parent in the layer list).
+  STREETS_GROUP = 'Move-In Streets',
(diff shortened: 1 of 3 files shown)


danielgoldbergtamu (approved):
Thanks.

PR #868, opened 2026-07-09
fix(aggiemap-angular) Use $gishost on new maps

sloneearly:
This PR moves the following maps that were previoiusly on dev to prod using $gishost:

Fish Camp, Fire School, Beef Cattle, Games of Texas


danielgoldbergtamu (approved):
Thanks.

PR #890, opened 2026-08-04
fix(aggiemap-angular) Turn on all Move-In layers by default

### LINKED ISSUE #887: aggiemap-angular: Move In needs all layers on by default
### Map loads in with lots and drop off zones only, would be difficult to understand on desktop or mobile
Currently opens like this:
<br>
Image

<br><br>

Map should have all layers on by default and look like this:
<br>
Image

sloneearly:
This PR enables all layers on the Move-In map by default.

Move-In map when first opened:

Screenshot 2026-08-04 080220

Closes #887


danielgoldbergtamu (approved):
Thanks!

PR #904, opened 2026-08-06
feat(aggiemap-angular): Fish camp layer updates

### LINKED ISSUE #869: aggiemap-angular: Fish Camp Map Updates
### Issue 1: Add Session G to builder
Session G needs to go with A & D for the lot 40 map

### Issue 2: Only one layer selectable/on at a time
Either send off for camp (which needs to be on by default) or bring back

sloneearly:
A bit late now, but this PR updates the following for Fish Camp:

- Session G now goes with A & D for the lot 40 map

image
image

- Send off is on by default
- Only one layer can now be turned on at a time

https://github.com/user-attachments/assets/bdac68a3-9430-4ada-a893-435fca2a0964

Closes #869


danielgoldbergtamu (approved):
Thanks.

PR #918, opened 2026-08-10
fix(aggiemap-angular): Remove unnecessary RV Lot labels from Main Campus Map

### LINKED ISSUE #917: aggiemap-angular: remove Reed Lots RV space numbers from Campus Main Parking Map
### Only certain lots have RV space numbers 
<br><br>
They are: Aggie RV, Lot 100d, Lot 58, and Lot 100j (south corner only). 

Image
<br><br>
There should not be RV numbers or any numbers visible in 100e, what it looks like on Aggie Map:
<br>

Image

<br><br>

What it should look like;
<br>

Image

sloneearly:
This PR removes the RV lots from the Main Campus map while also keeping them on the actual RV lots.

image

Also keeps them where they're supposed to go, i.e. Lot 100d

image

Closes #917


danielgoldbergtamu (approved):
Gross that we have to do it this way, but the data are like this, so there is no other way.

PR #931, opened 2026-08-26
fix(aggiemap-angular) Fix for icons not showing on sport maps

sloneearly:
This PR fixes an issue where icons on sport maps did not display and/or were replaced with bullet point icons.

 - Baseball
image

- Indoor track
image

- Men's Basketball
image

- Outdoor Track
image

- Soccer
image

- Softball

image

- Swimming
image

- Volleyball
image

- Women's Basketball
image

Closes #390


sloneearly:
Not entirely sure if this is what the crosswalk icon is supposed to be. I don't have access to view the new layer sources. If this is not what they're supposed to look like, we can fix it fairly easily.

image

danielgoldbergtamu (changes requested):
I do not like that it created a new class for athletic symbols instead of keeping the fixes contained within each of the specific events. Can you re-run it again and ask it to keep the symbols contained in and specific to each event?

sloneearly:
@danielgoldbergtamu All the symbols should be back in their respective events now

danielgoldbergtamu (approved):
Thanks, I know that we should be centralizing the symbols so we don't have to keep them duplicated all over, but would be a fix for the actual issue - that the server is not rendering properly, so I don't want to create a fix that is not the fix.

PR #933, opened 2026-09-02
fix(aggiemap-angular): Shared links to lots now drill down to lots

### LINKED ISSUE #840: aggiemap-angular: lot-specific copy-URL in event with builder does not link to lot
### Issue

For maps that have a builder, the copied link does not drill down to its respective lot. The copied link takes the user to the builder screen.

For example: For move-in:

Image

The link to a lot takes the user to the builder: https://dev.aggiemap.tamu.edu/parking/move-in/map/d?feature=Move-In+Lots%3A34815

### Resolution

1. This is probably occurring because all of the selections that a user did in the builder are not included on the copy-link, so when the application tries to load, it does not have the builder selections and doesn't know what to do.
2. In events that have builders, add the builder selection variables onto the copy URL so they can be processed when the link is used.

sloneearly:
This PR fixes an issue where shared links to different lots did not drill down to the lot itself. This PR makes shared links zoom to the respective lot.

image

Closes #840

### DIFF
--- a/libs/ts/events/ngx/src/lib/guards/settings/settings.guard.ts
+++ b/libs/ts/events/ngx/src/lib/guards/settings/settings.guard.ts
@@ -34,7 +34,11 @@ export class SettingsGuard implements CanActivate {
 
     try {
       // Call event settings service to update and set/overwrite any settings in local storage.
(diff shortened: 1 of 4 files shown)


danielgoldbergtamu (approved):
Thanks.

PR #939, opened 2026-09-14
fix(aggiemap-angular) Street Symbology check + T Camp and baseball fix

### LINKED ISSUE #924: aggiempa-angular: street closure symbology inconsistent
### Issue

There is a problem with how the server is rendering the street closure layers across various maps. It is supposed to be hatches, but is showing as solid red. For example on October Ring Day, Day 2.

Image

### Resolution

Update the street closure symbology across all maps to use the custom build renderer. Follow what was done on move in:

Image

sloneearly:
The issue outlined in #924 seems to have been mostly resolved already. All the symbology on both the dev and prod servers match. However, on the T camp map had one issue where its layer icons on the legend returned errors on prod and dev:

image

This has been fixed:

image

For baseball, the hatched circles aren't baseball closures. The layer has 90 features and only 3 are baseball's (the strip by Reed Arena and the shape near Penberthy). 64 of the rest are other sports' gate circles. Prod has never shown them. As such, the gate layer now filters to baseball = 1.

Closes #924


danielgoldbergtamu (approved):
looks good.

